Issue #48, opened 2015-02-21
Allow datetime objects in TinyDB


EmilStenstrom:
Since JSON is the default serilization format for TinyDB there's the datetime problem:

```python
>>> from tinydb import TinyDB
>>> from datetime import datetime
>>> db = TinyDB("db.json")
>>> db.insert({"date": datetime.now()})
...
TypeError: datetime.datetime(2015, 2, 21, 17, 24, 17, 828569) is not JSON serializable
```

Many other databases handle datetime conversion for the user, and I would very much like TinyDB to do the same (It's usually fixed by specifying a [custom encoding](http://stackoverflow.com/questions/10721409/why-does-json-serialization-of-datetime-objects-in-python-not-work-out-of-the-bo) and a corresponding decoder when reading from the database).

Do you think this is a good idea?

msiemens:
I defenitely think it's useful but I'm not sure how to implement this best. I wouldn't make this a default, though, because that would break usages where a conversion is not expected. Instead one could make use of TinyDB's extensibility. At the moment I can think of these options:

- Create a custom storage (copying the source of `JSONStorage`) that uses a custom JSON encoder/decoder
- Implement the encoder/decoder as a middleware. It would scan the data for `datetime` objects when writing and for date-like strings when reading and convert between these two

EmilStenstrom:
Thanks for your quick reply!

Since it's not possible to insert datetime objects right now, adding an encoder wouldn't break any existing usages. There could be issues decoding, but only if the encoded format clashes with something that someone already uses. Maybe a string like "TinyDate(2015-01-01T01:01:01)" would be sufficiently unique to make sure existing applications continue working.

I think this is something that's worthwhile to have in "core" since it allows simple date fetching methods to be added to TinyDB. Adding methods like: `db.search(month("timestamp") == 2)` looks like a great fit for TinyDB I think, and as far as I understand you would need to control the datetime serialization to get that to work.

Meanwhile, I think I will do the encoding/decoding in the application layer.

msiemens:
> I think this is something that's worthwhile to have in "core"

Basically this is a design question. I agree that this feature will be useful but from a design perspective I would prefer this to be an extension (something like `tinydb-datetime`). The main reason is that IMO the core should be kept as small as possible while offering comprehensive support for extensions of all kinds.

Maybe we could add some kind of support for custom encoders/decoders to TinyDB so users can use it without hassle. This could look like this:

```python
class DatetimeSerializer(Serializer):
[...]

PR #50, opened 2015-03-17
Add support for custom serializers

msiemens:
As discussed in #48, it would be nice to have support for custom serialization of objects unknown to TinyDB (= not supported by the JSON backend). This implements this interface, but I'm unsure how useful it will be.

cc @eugene-eeo @EmilStenstrom

### DIFF
--- a/docs/extend.rst
+++ b/docs/extend.rst
@@ -22,12 +22,17 @@ for ``datetime`` objects could look like:
         def decode(self, s):
             return datetime.strptime(s, '%Y-%m-%dT%H:%M:%S')
 
-We can use this Serializer like this:
+To use the new serializer, we need to use the serialization middleware:
 
 .. code-block:: python
 
-    >>> db = TinyDB('db.json')
-    >>> db.register_serializer(DateTimeSerializer(), 'TinyDate')
+    >>> from tinydb.storages import JSONStorage
+    >>> from tinydb.middlewares import SerializationMiddleware
+    >>>
+    >>> serialization = SerializationMiddleware()
+    >>> serialization.register_serializer(DateTimeSerializer(), 'TinyDate')
+    >>>
+    >>> db = TinyDB('db.json', storage=serialization)
     >>> db.insert({'date': datetime(2000, 1, 1, 12, 0, 0)})
     >>> db.all()
     [{'date': datetime.datetime(2000, 1, 1, 12, 0)}]
--- a/docs/extend.rst
+++ b/docs/extend.rst
@@ -1,6 +1,38 @@
 How to Extend TinyDB
 ====================
 
+Write a Serializer
+------------------
+
+TinyDB's default storage is fairly limited when it comes to supported data types.
+If you need more flexibility, you can implement a Serializer. This allows TinyDB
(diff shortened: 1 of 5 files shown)


EmilStenstrom:
This looks excellent to me. Will use for all projects where I want to store dates in the database.

eugene-eeo:
I remain skeptical but I think there must be a better way of doing this... Currently the code looks quite complex and no offence, "hack-ish" but I'm sure you'd clean it up. Coming from a performance point of view, I think the registered serializers should be stored in a hash-table, i.e.

```python
self._serializers = {}
```

Then prefix lookup will be very efficient and straightforward. And it also forces developers to choose a unique name ;)

msiemens:
First, sorry for not moving forward with this PR for so long! I hope we're now ready to go.

@eugene-eeo I've now refactored the serialization/deserialization into methods. Now I also use a dict for storing the serializers.

> Also better to move the serializing and deserializing parts into a separate class, then add hooks in order to preserve current performance since not every use case demands custom serialization.

I don't think creating a hooking infracstructure will perform better than the current code because when no serializers are registered as the outer for-loop of the (de)serialization will not run at all.

But on the other hand I'm currently toying with the idea of using a middleware for serialization. This could be used like this:

```python
serializer = Serializer(JSONStorage)
serializer.register(DateTimeSerializer(), 'TinyDate')
db = TinyDB(storage=serializer)

# ...
```

The advantage of this approach is that it doesn't modify the core at all and is totally opt-in. What do you think, @EmilStenstrom, @eugene-eeo?

eugene-eeo:
:+1: for not modifying core. By separating hooking infrastructure I meant the same thing as keeping it out of core and attaching the hooking class as a middleware, sorry for not explaining myself clearly.

msiemens:
@eugene-eeo I've updated this PR again, now with a Middleware that does the serialization. The only downside is that its usage is now less ergonomic (see [` docs/extend.rst `](https://github.com/msiemens/tinydb/commit/c36f25cea8f379bf6dba14f72f553e75c3f4dc23?short_path=602dd9c#diff-602dd9ca99aff027edb5d0fd06a9be6c)).

eugene-eeo:
Just a suggestion, I think I must be frustrating you but why not make it an extension? Custom serialisation is not something that we'd expect to be in core.

EmilStenstrom:
As I've said before, I'm in favor of having this in core. Having dates in JSON is very common in my experience, and being able to point to the serialization docs is a big advantage over having to install an extension in my opinion.

msiemens:
I agree with @EmilStenstrom about the extension question. I think it's okay to ship the middleware with tinydb as it's totally opt-in and we already provide a middleware that could be moved to an extension (`CachingMiddleware`).

msiemens:
I agree with @EmilStenstrom about the extension question. I think it's okay to ship the middleware with TinyDB as it's totally opt-in and we already provide a middleware that could be moved to an extension (`CachingMiddleware`).

msiemens:
For the record: This feature has been moved to [tinydb-serialization](https://github.com/msiemens/tinydb-serialization) in [TinyDB 3.0](https://github.com/msiemens/tinydb/blob/master/docs/changelog.rst#v300-2015-11-13)

PR #54, opened 2015-05-15
If there is no modified data, do not write.

rainwoodman:
This avoids an exception:

    Exception ValueError: 'I/O operation on closed file' in <bound method CachingMiddleware.__del__ of  <tinydb.middlewares.CachingMiddleware object at 0x7f9c96610a90>> ignored

The case happens when the objects has been closed and __del__ is invoked. Since the semantics guarantees
no writes after close, avoid storage.write when there are no changes in flush seems to be a reasonable fix.

### DIFF
--- a/tinydb/middlewares.py
+++ b/tinydb/middlewares.py
@@ -111,9 +111,9 @@ class CachingMiddleware(Middleware):
         """
         Flush all unwritten data to disk.
         """
-
-        self.storage.write(self.cache)
-        self._cache_modified_count = 0
+        if self._cache_modified_count > 0:
+            self.storage.write(self.cache)
+            self._cache_modified_count = 0
 
     def close(self):
         self.flush()


msiemens:
Thank you for the PR! Do you have an idea how this could be tested?

rainwoodman:
This would do it. I am not sure how to integrate into the test framework. Shall I put this in test_middlewares.py?


```
from tinydb import TinyDB
from tinydb import where
from tinydb.storages import JSONStorage

class DB(TinyDB):
    def __init__(self, filename):
        TinyDB.__init__(self, filename, storage=CachingMiddleware(JSONStorage))
        self.foo = filename

def main():
    with DB("abcd.json") as db:
        for i in range(12):
            db.remove(where("key") == i)
            d = {'key':i,
                 'value':"1231",
                }
            db.insert(d)

main()
```

msiemens:
Hmmm... The ignored exception is raised when garbage collecting the database object while the file handle has already been closed. It appears to me that on garbage collection the file handle is collected (and closed) first, followed by the storage/middleware. But this of course depends on the Python implementation and maybe even version¹, so I'll merge this PR without an accompanying test.

¹ I've tried to reliably force garbage collection of the database object, but haven't succeeded so far.

msiemens:
I've now released v2.3.2 which contains your changes :smiley:

rainwoodman:
Thanks!

PR #60, opened 2015-06-03
JSONPrettyStorage for pretty printed storage.

rainwoodman:
This is not pretty. I wrapped JSONStorage class in a factory function for this.
Similar pattern is for the CachingStorageMiddleware The number of cached operations can't be provided as a parameter and may need to be modified as well. 

I am happy to see a more elegant alternative!

### DIFF
--- a/tinydb/storages.py
+++ b/tinydb/storages.py
@@ -64,43 +64,47 @@ class Storage(with_metaclass(ABCMeta, object)):
 
         pass
 
-
-class JSONStorage(Storage):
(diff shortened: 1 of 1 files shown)


msiemens:
Thanks for the pull request! While the JSON storage could use more flexibility, I think the implementation can be simpler. I've implemented an alternative in #61. Could you have a look at it?

rainwoodman:
Closed as dup of #61

PR #61, opened 2015-06-03
Allow users to specify additional arguments for the JSON storage

msiemens:
Alternative implementation for #60. Example usage:

```
db = TinyDB('db.json', sort_keys=True, indent=4, separators=(',', ': '))
```

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -384,6 +384,13 @@ To use the in-memory storage, use:
 >>> from tinydb.storages import MemoryStorage
 >>> db = TinyDB(storage=MemoryStorage)
 
+.. hint::
+    All arguments except for the ``storage`` argument are forwarded to the
+    underlying storage. For the JSON storage you can use this to pass
+    additional keyword arguments to Python's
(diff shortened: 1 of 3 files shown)


rainwoodman:
While this works, I worry in the future we can no longer tell if a parameter is for JSONStorage or for TinyDB itself.

What about putting these storage arguments into a single argument?

```
db = TinyDB('db.json', storage_args=dict(sort_keys=True, indent=4, separators=(',', ': ')))
```

(but the name storage_args looks ugly.)

msiemens:
> While this works, I worry in the future we can no longer tell if a parameter is for JSONStorage or for TinyDB itself.

That's a good point. But right now the documented behaviour is that all arguments to the constructor except for `storage` are passed to the underlying storage. So basically, if we want to add arguments for TinyDB we're in trouble anyhow (the current way of "dealing" with this is by adding keyword arguments to `Tinydb.table` – not pretty but better than nothing).

rainwoodman:
I see! Your original approach more consistent than my suggestion. I have no objections to merging this (and closed #60 as well). 

Thanks.

Issue #63, opened 2015-06-22
Allow for additional arguments in custom test functions


lordjabez:
I'm finding myself using custom test functions a lot, many of which are nearly identical, and writing the same lambda function over and over is getting tedious.

It'd be really handy if one could pass additional arguments to these functions, so that one could do this:

```
def case_insensitive_match(a, b):
    return a.casefold() == b.casefold()

db.where('value').test(case_insensitive_match, my_val)
db.where('value').test(case_insensitive_match, my_other_val)
```

If you're keen on having this ability too, I'd be happy to code it up and submit a pull request.

P.S. Just discovered this library, and love it so far!

msiemens:
That sound like a nice addition! I'll gladly accept the pull request if you submit one :)

lordjabez:
Cool. I'll work on it this weekend.

robintw:
Is it possible to release a new version to PyPI including this change? I'm writing another module that depends on this functionality, and it'd make things a lot easier for my users to be able to just install from PyPI.

lordjabez:
That'd be up to @msiemens.

msiemens:
Hey all, sorry for letting this wait for so long! I've been quite busy lately with university, but I've now released version 2.4 of TinyDB which includes this feature. I'll try to catch up with the other issues next week.

PR #64, opened 2015-06-29
Allow parameters in custom test functions

lordjabez:
This is to address issue #63.

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -112,6 +112,18 @@ def test_custom():
     assert not query({'': None})
 
 
+def test_custom_with_params():
+    def test(value, minimum, maximum):
+        return minimum <= value <= maximum
+
+    query = where('val').test(test, 1, 10)
+
+    assert query({'val': 5})
+    assert not query({'val': 0})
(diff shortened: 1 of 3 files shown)


tests/test_queries.py
```diff
@@ -112,6 +112,19 @@ def test(value):
...
+def test_custom_with_params():
+    def test(value, minimum, maximum):
+        return minimum <= value <= maximum
+
+    query = where('val').test(test, 1, 10)
+
+    assert query({'val': 5})
+    assert not query({'val': 0})
+    assert not query({'val': 11})
+    assert not query({'val': '44'})
```
msiemens: You're using a string for `44`, not an int. This breaks the tests for Python3 and PyPy3 (see https://travis-ci.org/msiemens/tinydb/builds/68895961).

msiemens:
There's a tiny mistake in a test, but otherwise it looks good!

lordjabez:
Should I just remove that test? I only included the string comparison because there was a similar one for the previous test case.

lordjabez:
Ahh, I see the issue now. I'll simply remove the test.

Issue #67, opened 2015-08-13
Reading an existing .json file written out using TinyDB


tomwallis:
Apologies if this is a stupid question, but I have not seen the answer in your docs.

A colleague has passed me a `.json` file written out using TinyDB. I would like to read its contents. However, if I use

```
db = TinyDB('my_file.json')
```

then `db` is empty, and the file on disk is overwritten (contains one line `{"_default": {}}`). What's the correct way to read this file into a Python interpreter? I see no "load" method in the docs and `tinydb.Storage.read` doesn't do this.

eugene-eeo:
Slightly weird question but is the JSON file valid? The last time I worked on the code we used a `try..except` block to test if files were empty, so if the file was manually modified and some minor issues were present (think missing quotes, unbalanced parens etc) then it would result in the same case as an empty file. So try loading it manually using `json.loads` to see if it works.

tomwallis:
Ah, you seem to have it right. 

```import json
with open('tom_mean_feature_activations.json', 'r') as f:
    data = json.load(f)

...


/Users/user/miniconda3/envs/default/lib/python3.4/json/decoder.py in raw_decode(self, s, idx)
    357         """
    358         try:
--> 359             obj, end = self.scan_once(s, idx)
    360         except StopIteration as err:
    361             raise ValueError(errmsg("Expecting value", s, err.value)) from None

ValueError: Expecting ',' delimiter: line 1 column 12582912 (char 12582911)

```

The `.json` file my colleague gave me appears to be invalid. Issue solved, but I suppose it would be helpful if tinyDB reported an error rather than silently overwriting the file.

msiemens:
Actually I'd like to keep this open. TinyDB currently assumes that if the JSON module raises a `ValueError`, the database file was empty. But as in this case, that's a pretty shaky assumption and something I want to revisit for a potential v3.0.

eugene-eeo:
Unhelpful but encouraging :+1: for looking into this issue.

msiemens:
This should be closed in the 3.0 release, I forgot to close this issue.

PR #69, opened 2015-09-17
[WIP] V3.0

### LINKED ISSUE #62: Refactoring Query objects
This is a proposal for a much simpler Query model that can be used, which essentially enables us to do the following, even without any advanced metaclass or inheritance magic:

```python
User = Query()
db.insert({'metadata': {'id': 1}})
db.find(User.metadata.id == 1)
```

Which is very SQLAlchemy-like and what one would expect, and I would say even more intuitive than the current style/model. The following is a simple implementation that only handles the very basic:

```python
import operator

class Query(object):
    def __init__(self, function=None):
        self.function = function or (lambda k: k)

    def __getattr__(self, attr):
        return Query(operator.itemgetter(attr))

    def __eq__(self, value):
        return Query(lambda k: self.validate(k) == value)

    def validate(self, value):
        return self.function(value)
```

The advantages over the current style is perhaps that this is much more intuitive and simple. We don't need to have a new class for every special case. It is also simpler to generate the intermediate query expressions so that they can be cached, instead of the rather primitive or hacky string representations that we are having at the moment, which is error prone and needs us to worry about Unicode etc.

msiemens:
This PR prepares an overhaul of TinyDB released as version 3.0.

Notable changes:

- New Query model:
    - `where('...').contains('...')` has been renamed to `where('...').search('...')` to indicate that it works on regexes
    - Support for ORM-like usage: `User = Query(); db.find(User.name == 'John')`
    - `where('foo')` is an alias for `Query().foo`
    - `where('foo').has('bar')` is replaced by either `where('foo').bar` or `Query().foo.bar`
        - In case the key is not a valid Python identifier, array notation can be used: `where('a.b.c')` is now `Query()['a.b.c']`
    - Checking for the existence of a key has to be done explicitely: `where('foo').exists()`
- Migrations from v1 to v2 have been removed
- `SmartCacheTable` has been moved to [msiemens/tinydb-smartcache](https://github.com/msiemens/tinydb-smartcache)
- Serialization has been moved to [msiemens/tinydb-serialization](https://github.com/msiemens/tinydb-serialization)

Todo:

- [x] Revise Query concept (see #62)
- [x] Revise storage concept (see #67, #70)
- [x] ~~Revise~~ Document extension concept (storages, middlewares, custom table classes)
- [x] Move `SmartCacheTable` to an extension package (depends on msiemens/tinydb-smartcache#1)
- [x] Move serialization to an extension package
- [x] List new extensions in docs
- [x] Remove migrations (which were for v1→v2)
- [x] Update the documentation
- [x] Update version number & release date

Closes #62, #67, #70


msiemens:
I've finally gotten around to finish the missing bits.

@eugene-eeo Could you have a look at this whether we're missing something for v3.0?

eugene-eeo:
@msiemens you probably want to expose some API for clearing the cache "properly"- at the moment tinyrecord relies on the ``_cache`` dictionary. I especially like the new Query implementation and how it looks more like an ORM while probably performing better and having less code to maintain than the old model. :+1:

msiemens:
> you probably want to expose some API for clearing the cache "properly"

Good idea, I've added `Table.clear_cache`.



> A little critique but I would probably use a frozenset to represent the __and__ and __or__ "nodes" just so that it is more correct- A AND B is the same as B AND A, and the same applies to A OR B, mathematically.

Good point! I've updated the AND and OR operations to be commutative.

As always, thanks for your insight, @eugene-eeo!

msiemens:
TinyDB v3.0.0 is now officially released :tada:

PR #70, opened 2015-09-25
Fix exception AttributeError

petrsnm:
Bug can be reproduced with the following snippet:

```
from tinydb import TinyDB, where
try:
    db = TinyDB("/does/not/exist/db.json")
except FileNotFoundError:
    # do something else
    pass
```

Results in: 
```
Exception ignored in: <bound method JSONStorage.__del__ of <tinydb.storages.JSONStorage object at 0x7f0b27a7ea90>>
Traceback (most recent call last):
  File "python/lib/python3.4/site-packages/tinydb/storages.py", line 94, in __del__
    self.close()
  File "python/lib/python3.4/site-packages/tinydb/storages.py", line 91, in close

    self._handle.close()
AttributeError: 'JSONStorage' object has no attribute '_handle'
```

### DIFF
--- a/tinydb/storages.py
+++ b/tinydb/storages.py
@@ -87,7 +87,8 @@ class JSONStorage(Storage):
         self._handle = open(path, 'r+')
 
     def close(self):
-        self._handle.close()
+        if hasattr(self, '_handle'):
+            self._handle.close()
 
     def __del__(self):
         self.close()


krisztianfekete:
This is a perfect issue, but unfortunately a problematic pull request.

Problems I see with this fix:
- Creates a solution to a problem that is not obvious:
  `JSONStorage.__init__()` does seems to create `self._handle`, but is not due to an exception, and there is a `__del__()` that implicitly uses it on a not fully initialized object.
- There are no unit tests (regression tests) showing the problem with the unpatched code and working with the fix - it is important as neither the problem nor the solution is trivial.
- Uses `hasattr()` which is usually easily avoidable  (this was the first sign of problems with the patch, and thus the one that triggered my further investigation).

Analysis of the problem:

`open()` in `touch()` fails due to non-existing directory, the irony is that `touch()` is probably supposed to be guarding against exactly this problem, but it turns out to be an easy but bad solution.

The existence of `__del__()` is a problem in itself: it suggests that implicit cleanup and reference counting as the garbage collector is depended upon by `TinyDB`!   Can it run out of file descriptors with PyPy, Jython?

It is not possible to call `close()` without `self._handle`, except from `__del__()`, thus the fix either should go to `__del__()` (workaround) or `TinyDB` should be changed not to depend on `JSONStorage.__del__()` (proper fix).

I suggest opening an issue about it and questioning the existence of `JSONStorage.__del__()`.

msiemens:
When writing the original code I didn't expect that Python calls the cleanup on objects where `__init__` failed with an exception ([details on StackOverflow](http://stackoverflow.com/a/6409718/997063)). Hence I assumed that `_handle` exists when cleaning up which results in the reported bug.

> open() in touch() fails due to non-existing directory, the irony is that touch() is probably supposed to be guarding against exactly this problem, but it turns out to be an easy but bad solution.

Actually `touch()` is used because the file is opened with `r+` mode which doesn't create a new file if it doesn't exist. It's fine that an exception is raised when the directory doesn't exist but it's not fine that cleanup fails in this case.

> I suggest opening an issue about it and questioning the existence of JSONStorage.__del__().

The more interesting issue here is how to do cleanup on storages. At the moment the concept isn't really clear. We have `db.close()` which calls `close()` on the storage. But users will probably forget to call it, so there is 1) a context manager which calls `close()` for you, and 2) JSONStorage implements `__del__` which calls `close()`. This of course may lead to bugs if the storage expects `close()` to be called just once. I'd be happy to revisit this topic for [v3.0](https://github.com/msiemens/tinydb/pull/69), but I'm not sure how to solve this in a elegant and solid way. What are your thoughts on this, @krisztianfekete?

krisztianfekete:
@msiemens it is hard what you ask as I do not really know `TinyDB`, so do not know how much my ideas would break code/upset users.

Anyway, here are my thoughts in the order of least noticeable to most radical:

----
I'd get rid of `__del__` as it is problematic and is a fallback only for the case when calling `close()` is forgotten.

----
I'd get rid of `TinyDB.close()` and provide a context manager only interface, thus forcing users to use it properly with `with` statements: `__enter__` could return a value that provides the `TinyDB` API, but the returned object's class is not public.

something like
```python
@contextlib.contextmanager
def TinyDB(*args, **kwargs):
    # create and manage private storage here
    storage = get_storage(*args, **kwargs)
    db = _TinyDB(storage)
    try:
        yield db
    finally:
        storage.close()

class _TinyDB(object):  # ~ as current TinyDB, but not exported
    ...
```
The context manager interface and `close` on `Table` is not needed.

----
`TinyDB` and `Table` has a strange relation: `TinyDB` is_a `Table` in a duck typing sense.
I assume this is because there are users who are OK with only one table, or even it is the most common use case.
So I'd reverse the relationship - provide a table centered API from the start and make the database object an implicit internal detail of tables.
- `tinydb_open(..., name='_default')`  -> context manager for the database, returning a `Table`, to be used like this:

    ```python
    with tinydb.open(..., name='_default') as table:
        main_program(table)
    ```
- new `Table` methods that are currently on `TinyDB` for advanced users with multiple tables:
  - `Table.siblings()` -> `set(Table names)`
  - `Table.sibling(name)` -> `Table`
  - `Table.purge_siblings()`

msiemens:
@krisztianfekete Thank you for your input, that helped me a lot. After putting some thought and research into this, I've come up to the conclusion that the main problem is that it's not clear who is reponsible for cleaning up (the database object? the table? the storage itself?). I think it's best if the database object is reponsible for calling `close()` on its storage at most once. This may happen when the user calls `close()` manually or on garbage collection. This is the same behaviour as implemented by Python's `sqlite3` and PyMongo. Also `close()` is only called if the storage has been initialized successfully. This will solve the bug found by @petrsnm. (Implementation in b79ce35addbdcd2d368e710d7149d009eafd7834).

petrsnm:
I agree.  This is actually what is thought when contemplating the pull
request that I sent to you.  However, I didn't feel qualified to submit
anything more than the simple fix and hope that you considered exposing a
close() method.

Thank you for looking at this carefully

PR #76, opened 2015-12-06
Minor grammar fix and duplicate text removal

maxtheman:
May need to update https://tinydb.readthedocs.org/en/latest/intro.html#why-not-using-tinydb based off of this -- not sure how read the docs works!


msiemens:
Thanks, @maxtheman!

PR #81, opened 2015-12-16
Adding GzipJSONStorage

MatthewTurk:
This adds a GzipJSONStorage class.  As near as I can tell, the Gzip module doesn't support simultaneous read/write, so it doesn't reuse the handle between different calls to read/write.  This is particularly helpful for our use case where we have many records that share attributes; using gzip shows an enormous reduction is size for the total data size.  The performance is worse, as expected.

### DIFF
--- a/tests/test_storages.py
+++ b/tests/test_storages.py
@@ -6,7 +6,8 @@ import pytest
 random.seed()
 
 from tinydb import TinyDB, where
-from tinydb.storages import JSONStorage, MemoryStorage, Storage
+from tinydb.storages import JSONStorage, MemoryStorage, Storage, \
+    GzipJSONStorage
 
 element = {'none': [None, None], 'int': 42, 'float': 3.1415899999999999,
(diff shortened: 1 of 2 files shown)


msiemens:
Thanks for the PR @MatthewTurk. Do you know why the tests failed on Python 3 and PyPy2/3? (See https://travis-ci.org/msiemens/tinydb/builds/97332270)

msiemens:
On a more general note, I think this storage would be better suited for a TinyDB extension as I want to keep the TinyDB core as minimal as possible. If you release this storage as an extension, I'll link to it in the docs and the readme. What do you think?

divanovGH:
http://stackoverflow.com/questions/5471158/typeerror-str-does-not-support-the-buffer-interface

msiemens:
@MatthewTurk What do you think about releasing a GzipJSONStorage as an extension to TinyDB?

MatthewTurk:
Great idea -- I'll work on it as such a thing.

PR #82, opened 2015-12-20
Add pickle storage

divanovGH:
Native objects (like datetime) serialize support

### DIFF
--- a/tests/test_storages.py
+++ b/tests/test_storages.py
@@ -6,7 +6,7 @@ import pytest
 random.seed()
 
 from tinydb import TinyDB, where
-from tinydb.storages import JSONStorage, MemoryStorage, Storage
+from tinydb.storages import PickleStorage, JSONStorage, MemoryStorage, Storage
 
 element = {'none': [None, None], 'int': 42, 'float': 3.1415899999999999,
            'list': ['LITE', 'RES_ACID', 'SUS_DEXT'],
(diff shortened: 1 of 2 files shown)


msiemens:
Thanks for the PR @divanovGH. I've already played with the idea of adding a picke storage. In the end I decided against it, because allows arbitrary code execution. The docs write:

> The `pickle` module is not secure against erroneous or maliciously constructed data. Never unpickle data received from an untrusted or unauthenticated source.

For that reason I'm uncomfortable putting including this in the TinyDB core. What do you think on that topic?

divanovGH:
I think that this solution can be used if the database is exclusively on the local machine.

divanovGH:
I think that this solution can be used if the database is exclusively on the local machine.

divanovGH:
this storage should be used at user's own risk

eugene-eeo:
@msiemens note that, though pickle *does* allow for arbitrary code execution, we are only using the pickle module to store "trusted" data - and it is a suitable tool in that respect. So as long as you don't simply open a database downloaded from somewhere else using the pickle storage, you are safe. Side note: the pickle storage should ideally live as an extension module.

msiemens:
> @msiemens note that, though pickle does allow for arbitrary code execution, we are only using the pickle module to store "trusted" data - and it is a suitable tool in that respect. So as long as you don't simply open a database downloaded from somewhere else using the pickle storage, you are safe.

This presumes the assumption that the file system is a "trusted" store and this isn't a really safe assumption. While I don't expect that anyone would target a pickle file used by TinyDB, I don't feel good including this in the TinyDB core. I'd really prefer if this is an extension module.

msiemens:
@divanovGH What do you think about releasing a pickle storage as an extension to TinyDB?

divanovGH:
@msiemens I completely agree with this

msiemens:
Okay, I'll then close this PR and if someone creates a TinyDB extension providing a pickle storage I'll gladly link to it from the docs.

PR #94, opened 2016-02-14
Min max feature

fAntel:
Hello
I added methods `min` and `max` to class Table which returns element with min/max value in specified field. I need it in my project and I want to share. What do you think about it?
PS I thought about add this methods to class Query for writing something like
`db.search((where('field1') == '1') & (where('field2').max()))`
but I couldn't figure out how to do this because value of `max` function is unknown in that moment when `generate_test` invoked. And I don't sure this is better than `db.max('field2', where('field') == '1'))`.

### DIFF
--- a/tests/test_tinydb.py
+++ b/tests/test_tinydb.py
@@ -420,3 +420,13 @@ def test_max(db):
     assert db.max('char', where('int') == 2)['char'] == 'f'
     assert db.max('char')['int'] == 2
     assert db.max('char')['char'] == 'f'
+
+def test_min(db):
+    assert db.min('char', where('int') == 1)['char'] == 'a'
+    assert db.min('char', where('int') == 2) is None
+    assert db.min('float', where('int') == 1) is None
+    db.insert_multiple({'int': 2, 'char': c} for c in 'def')
+    assert db.min('char', where('int') == 1)['char'] == 'a'
+    assert db.min('char', where('int') == 2)['char'] == 'd'
+    assert db.min('char')['int'] == 1
+    assert db.min('char')['char'] == 'a'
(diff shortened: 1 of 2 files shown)


eugene-eeo:
I don't mean to sound rude, but we actually don't need to add another method (and tests for edge cases) to the `Query` object to do this- simply,

```python
max(db.search(where(...)), key=lambda doc: doc["key"])
```

Or even better using the `operator` library:

```python
from operator import itemgetter
max(..., key=itemgetter("key"))
```

Another advantage is that the `max` function is more generic and works on most data structures in Python (lists, iterables, tuples).

msiemens:
> I added methods min and max to class Table which returns element with min/max value in specified field. I need it in my project and I want to share. What do you think about it?

Using the Table class for this feels wrong to me. Altough -- as you mentioned -- it currently can't be implemented using the Query class as it only selects elements based on the current value. This is also true for other queries that would act on the list of all matching queries (e.g. select every other item, select the n-th item etc.).

As far as I see, the solution @eugene-eeo proposed be the simplest one -- if you want to have the semantics of the logical AND operation. So would be equivalent:

```python
db.search((where('field1') == '1') & (where('field2').max()))
# and
max(db.search(where('field1') == '1'), key=lambda el: el["field2"])
```

On the other hand, you cannot use the proposal to express disjunction (logical OR):

```python
db.search((where('field1') == '1') | (where('field2').max()))
```

Interestingly, to express that in MySQL, you'd have to use the `GROUP BY` clause for min/max. It gets even messier (subqueries etc.) if you want to use other aggregate functions. And IMO we shouldn't replicate that complexity in TinyDB.

So yeah, I think min/max queries are beyond what's possible with TinyDB.

msiemens:
For your own use, you can use a [custom table class](http://tinydb.readthedocs.org/en/latest/extend.html#creating-a-custom-table-classes) to add that functionality to your instance of TinyDB while still benefitting from updates to the main package. That's propably the cleanest solution.

fAntel:
@eugene-eeo, thank you very much for your solution. I actually didn't notice that `max` can return exactly dict from list of dicts. This is much better than my solution

PR #98, opened 2016-04-20
Provide a way to specify a non-default table name when creating TinyD…

deemson:
…B instance.

### DIFF
--- a/tests/test_tinydb.py
+++ b/tests/test_tinydb.py
@@ -409,3 +409,10 @@ def test_gc(tmpdir):
     table.insert({'int': 13})
     assert len(table.search(where('int') == 13)) == 1
     assert table.all() == [{'something': 'else'}, {'int': 13}]
+
+
+def test_non_default_table():
+    db = TinyDB(storage=MemoryStorage)
+    assert [TinyDB.DEFAULT_TABLE] == list(db.tables())
+    db = TinyDB(storage=MemoryStorage, table='non-default')
+    assert ['non-default'] == list(db.tables())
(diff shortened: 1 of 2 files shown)


msiemens:
Looks reasonable, thank you! While we're at it, could you also add a short test in `tests/test_tinydb.py`?

deemson:
Sure.

deemson:
moved kwargs pops to the top of init because storage instances don't recognize 'table' keyword

msiemens:
I must have missed it in the first commit, but you introduced a new kwarg parameter to `Table.__init__()`, right? I'm conflicted on that because technically that's a breaking change (storages cannot use the `table` keyword anymore). But on the other hand I don't think this will break anyone's code as writing a storage that takes a a `table` kwarg isn't a good idea in the first place.

deemson:
I've introduced a new `TinyDB.__init()__` kwarg and it won't be passed to storage, it is being popped out of kwargs dictionary right before the creation of storage instance.

eugene-eeo:
May be a matter of taste but I think using a longer keyword argument like `default_table` would be preferable here as it better reflects the purpose of the parameter, and leaves less room for confusion in the future.

As an aside, I don't see why one would want to override the default table. Don't get me wrong, I like how you're handling it elegantly here, but I don't see why the application would care about the default table name.

deemson:
I cannot argue with this type of reasoning, this is a matter of taste. I just don't want to see "_default" keyword in my storage files. I find it ugly.

Agreed on the parameter naming, though.

msiemens:
`default_table` sounds good to me. The name of the default table indeed is a matter of taste and I see no problem with allowing to rename it.

msiemens:
Merged! I hope to get a new release of TinyDB ready by the end of this week. Thanks everyone!

PR #104, opened 2016-06-13
Update database.py

KeksKeksKeks:
Don't write db if _default is an empty Table.

If there are no documents in the _default table, TinyDB will write (and thus, re-serialize) the entire database.
This is because bool({}) == False, although the table does exist.

I have a JSON file under version control, which doesn't use the default table. Recently, I noticed that the file changes although I didn't change any document within the database. 
Apparently, TinyDB.__init__ reads the default table, which is empty in my case. An empty document evaluates to False in a boolean context, and TinyDB will write a new, empty table, effectively re-serializing the database on read.

### DIFF
--- a/tests/test_tinydb.py
+++ b/tests/test_tinydb.py
@@ -5,7 +5,7 @@ import pytest
 
 from tinydb import TinyDB, where
 from tinydb.storages import MemoryStorage
-
+from tinydb.middlewares import Middleware
 
 def test_purge(db):
     db.purge()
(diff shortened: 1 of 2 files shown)


KeksKeksKeks:
Classical case of RTFM :(

I should have read the contribution instructions first. I have added a test to demonstrate the original behaviour, introduced changes to make it go away, and added the neccessary changes to make the coverage check happy.

Kudos for such a thorough code infrastructure! I enjoyed using (and now, contributing to) TinyDB.

msiemens:
That makes sense, thank you! Just out of curiosity: I find it interesting that the VCS tracks the file modification time. Which one are you using?

KeksKeksKeks:
You're welcome!

I use git, which doesn't track file modification *time*. Instead, because python dicts are unordered, two JSON serialisations of the same database will look different. And a different file *content* is, of course, then detected by the VCS.

msiemens:
Ah, I see! Unexpected reordering of Python's dicts have bitten me more than once :smile:

eugene-eeo:
Turns out there is a `sort_keys` option in the `json.dumps`function. However I *think* that'd degrade performance slightly, but anyways here it is: http://stackoverflow.com/questions/10844064/items-in-json-object-are-out-of-order-using-json-dumps

PR #107, opened 2016-06-27
Rename Query.path to Query._path

joshuahhh:
I don't know how common it is to have a field called `path` in your database, but that's the situation I found myself in today, and I got mightily confused when `Query().path == 'some/path'` returned `False` instead of giving me a query object! It must be less likely to have a field called `_path`, so this should prevent some confusing collisions.

PTAL. Thanks!

### DIFF
--- a/tinydb/queries.py
+++ b/tinydb/queries.py
@@ -105,12 +105,12 @@ class Query(object):
 
     def __init__(self, path=None):
         if path is None:
-            self.path = []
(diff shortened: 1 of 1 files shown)


msiemens:
Thanks for the PR, @joshuahhh!

msiemens:
FYI, I've just released v3.2.1 which includes your fix :)

PR #112, opened 2016-08-06
Fix API docs: db.find() does not exist.

Fak3:
db.search() is the right name.


msiemens:
You're completely right, thanks for the PR!

PR #114, opened 2016-09-11
add create_dirs kwarg to JSONStorage

Ofekmeister:
Setting create_dirs=True in `__init__` will create the necessary parent directories to ensure db file will be created. This will further lessen what the user has to worry about.

### DIFF
--- a/tests/test_storages.py
+++ b/tests/test_storages.py
@@ -1,4 +1,6 @@
+import os
 import random
+import tempfile
 
 import pytest
 
@@ -69,6 +71,29 @@ def test_json_readwrite(tmpdir):
 
     db.remove(where('name') == 'A short one')
     assert get('A short one') is None
+
+
+def test_create_dirs():
+    temp_dir = tempfile.gettempdir()
+    db_dir = ''
+    db_file = ''
+
+    while True:
+        dname = os.path.join(temp_dir, str(random.getrandbits(20)))
+        if not os.path.exists(dname):
+            db_dir = dname
+            db_file = os.path.join(db_dir, 'db.json')
+            break
+
+    db_conn = JSONStorage(db_file, create_dirs=True)
+    db_conn.close()
+
+    db_exists = os.path.exists(db_file)
+
+    os.remove(db_file)
+    os.rmdir(db_dir)
+
+    assert db_exists
 
 
 def test_json_invalid_directory():
--- a/tinydb/storages.py
+++ b/tinydb/storages.py
@@ -15,7 +15,11 @@ except ImportError:
     import json
 
 
-def touch(fname, times=None):
+def touch(fname, times=None, create_dirs=False):
+    if create_dirs:
+        base_dir = os.path.dirname(fname)
+        if not os.path.exists(base_dir):
+            os.makedirs(base_dir)
     with open(fname, 'a'):
         os.utime(fname, times)
 
@@ -70,7 +74,7 @@ class JSONStorage(Storage):
     Store the data in a JSON file.
     """
 
-    def __init__(self, path, **kwargs):
+    def __init__(self, path, create_dirs=False, **kwargs):
         """
         Create a new instance.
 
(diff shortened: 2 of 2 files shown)


Ofekmeister:
@msiemens Python 3.2 is weird and and py.test is errorring.

eugene-eeo:
2 solutions I can think of:

 - pin to a specific version of `pytest`. Earlier the tests were not breaking on Python 3.2. travis already has `pytest` installed, so we probably need to remove [this line](https://github.com/msiemens/tinydb/blob/master/.travis.yml#L13) to see if travis installs a working pytest for us on 3.2.
 - drop 3.2 support altogether. It is more restrictive than 2.7 in terms of syntax (I think `b""` and `u""` is disallowed) and is the reason why we have to force ourselves to write strange, lowest denominator code.

Ofekmeister:
@eugene-eeo Thanks for the feedback!

Should the database handle intermediate directory creation? Perhaps not, but it would be a useful option. Consider the main use case for a local db: an app storing data in a designated place, usually cross-platform.

```python
# an_app/db.py

import os
from appdirs import AppDirs

DATA_DIR = AppDirs('app_name').user_data_dir
DB_FILE = os.path.join(DATA_DIR, 'db_name.db')

# You will see this idiom in every program using a local datastore
if not os.path.exists(DATA_DIR):
    os.makedirs(DATA_DIR)
```

tests/test_storages.py
```diff
@@ -71,6 +73,29 @@ def test_json_readwrite(tmpdir):
     assert get('A short one') is None
 
 
+def test_create_dirs():
+    temp_dir = tempfile.gettempdir()
+    db_dir = ''
+    db_file = ''
+
+    while True:
```
msiemens: I don't feel good about this loop. Couldn't this make tests hang for some edge cases? I'd propose using pytest's [tmpdir](http://doc.pytest.org/en/latest/tmpdir.html) that's also used in other tests (e.g. [test_storages.py#L49](https://github.com/msiemens/tinydb/blob/873a82e0b0ed71e6f2dcd3332e26960e95622675/tests/test_storages.py#L49)).

msiemens:
> Should it really be the job of the database module to handle intermediate mkdirs?

Well, the way it's implemented it's the job of the storage which that seams reasonable to me. I don't have a strong opinion for/against a change like this but still think it's a useful addition.

msiemens:
re Python 3.2: We don't support Python 3.0/3.1 either so I think it's fine to drop 3.2 too considering its usage is less than 1% (see pytest-dev/pytest#1627 for source + similar reasoning)

tests/test_storages.py
Ofekmeister: @msiemens It will never hang. The loop just generates a temporary directory with the assurance that it does not already exist.

msiemens: Using pytest's tmpdir would allow us to not have to retry directory creation. Anyways, it's not a big deal really.

Ofekmeister:
Would you like me to drop 3.2 in this PR?

msiemens:
I'll do that in a later commit. @eugene-eeo I suppose you're fine with dropping Python 3.2?

eugene-eeo:
@msiemens definitely. Python 3.2 for me is something that is 'midway' between 2 and 3. It doesn't have all the nice features of 3 but is also significantly different from 2.6+.

tests/test_storages.py
eugene-eeo: It probably will not hang. If you however used something like `getrandbytes` it may cause issues with valid path characters and such. Also you don't have to include the clean up though, I believe tmpdirs are [cleaned up automatically anyways](http://doc.pytest.org/en/latest/tmpdir.html#the-default-base-temporary-directory).

msiemens:
@eugene-eeo I'm a little confused, how does creating directories lead to corrupting user data?

eugene-eeo:
@msiemens sorry, I have phrased it badly. It wouldn't be corruption but more of storing (and reading) incorrect data as it masks errors from invalid paths.

Ofekmeister:
@eugene-eeo Can you please give an example?

eugene-eeo:
@Ofekmeister @msiemens now that I've thought about it, it's not really an issue and I've confused myself, sorry. As it stands I'm ok with creating intermediate directories, as long as it is an _explicit_ option.

msiemens:
Merged! Thanks everyone! I'll try to get a new release out sometime soon (this week or the next one)

PR #115, opened 2016-09-27
use predefined constant instand of magic number

yan97ao:


msiemens:
Merged, thank you! :)

PR #127, opened 2017-03-22
Update import from 'where' to 'Query'

eyalev:
'where' doesn't seems to be used in the following code, as opposed to 'Query'.


msiemens:
Nice catch, thank you!

PR #129, opened 2017-03-29
Proofread and correction of extend.rst

dutchovenbread:
I heard the Markus Siemens interview on the Talk Python podcast. He stated that he would like some editing of the documents for native-English correction. Most of these edits observe that "storage" and "middleware" are both non-count nouns in English (other non-count nouns are "data", "money," "weather", etc. Therefore, a native English speaker would never say, "a storage", "storages", "a middleware", or "middlewares". If you want to keep these terms in your documentation, they sound like neologisms. Unique, new, but not terrible. Not common usage.

Also, this is my first pull request, so I don't really know what happens with each button press here in GitHub. I am not sure that when I press "propose file change" if I will be able to modify the other files in the documentation under the same change. I will work on that tomorrow. Thanks for your understanding. If this modification is not what you had in mind, please comment as such.

### DIFF
--- a/docs/extend.rst
+++ b/docs/extend.rst
@@ -3,17 +3,17 @@ How to Extend TinyDB
 
 There are three main ways to extend TinyDB and modify its behaviour:
 
-1. custom storages,
-2. middlewares, and finally
+1. custom storage,
+2. custom middleware, and
 3. custom table classes.
 
 Let's look at them in this order.
 
-Write a Custom Storage
+Write Custom Storage
 ----------------------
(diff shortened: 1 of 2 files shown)


msiemens:
Hey @dutchovenbread, thanks for the Pull Request! I didn't even consider that these terms might be uncountable so thanks for pointing that out :)

> Also, this is my first pull request, so I don't really know what happens with each button press here in GitHub.

No worries, that's a perfect first pull request!

PR #139, opened 2017-05-26
Make table and db iterable

noirbizarre:
This PR makes `Table` iterable and `TinyDB` iterable over default table so it's possible to write:
```python
# Iter over default table
db = TinyDB()
for obj in db:
   do_something(obj)
table = db.table('my_table')
for obj in table:
   do_something(obj)
```


msiemens:
I think that's a nice addition. I'm still thinking about where this should be described in the docs: basic usage or advanced usage.

@eugene-eeo What do you think about this change?

noirbizarre:
I just added a simple documentation proposal

eugene-eeo:
@msiemens I'm fine with it since it opotimises for the 80% use case. But it needs to be documented somewhere. Perhaps in Basic Usage would be appropriate.

msiemens:
> since it opotimises for the 80% use case

Good point. I think it's quite intuitive for `iter(db)` to be an alias for `db.all()`. Thanks for the Pull Request @noirbizarre!

msiemens:
Also, I've now released TinyDB v3.3.0 which includes this feature :)

PR #140, opened 2017-05-26
Fix unclosed files warnings on Python 3+

noirbizarre:
This PR ensure opened files are closed during tests (and so removes the raised warnings on Python 3+)


msiemens:
Thanks!

PR #142, opened 2017-06-22
Remove absolute imports to support vendoring

mottosso:
Hi folks,

Thanks for this amazing project! I'm looking to bundle tinydb with another software library but couldn't due to the global references to `tinydb` in import statements.

This PR makes all imports relative, so that you can now do this.

```python
from my_library.vendor import tinydb
```

Let me know if there are any issues.

### DIFF
--- a/tinydb/__init__.py
+++ b/tinydb/__init__.py
@@ -8,8 +8,8 @@ backend. It has support for handy querying and tables.
 
 Usage example:
 
->>> from . import TinyDB, where
->>> from .storages import MemoryStorage
+>>> from tinydb. import TinyDB, where
+>>> from tinydb.storages import MemoryStorage
 >>> db = TinyDB(storage=MemoryStorage)
 >>> db.insert({'data': 5})  # Insert into '_default' table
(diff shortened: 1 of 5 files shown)


tinydb/__init__.py
```diff
@@ -8,8 +8,8 @@
 
 Usage example:
 
->>> from tinydb import TinyDB, where
->>> from tinydb.storages import MemoryStorage
+>>> from . import TinyDB, where
+>>> from .storages import MemoryStorage
```
msiemens: I think we should keep the absolute import here as that's what users will usually continue to use. Or am I mistaken?

msiemens:
One small nit with the docstring, otherwise looks good to me!

tinydb/__init__.py
mottosso: Oh, yes you are right. Sorry about that.

msiemens:
Merged, thanks! I'll try to find to release a new version of TinyDB next week :)

msiemens:
BTW, I've just released v3.3.1 which includes this pull request :)

PR #145, opened 2017-08-06
Added some trivial key/value update operations

pjrule:
I found the tinydb.operations functions to be a bit bare—there are cases where one might want to change a key by more than ±1, which is all the existing `increment()` and `decrement()` functions do. :) I also added a `set()` function to directly update keys.


msiemens:
That makes sense, merged :)

msiemens:
BTW: The new update operations are included in the latest release of TinyDB I just published (v3.4.0)! Thanks again for the contribution 🙂

PR #150, opened 2017-08-29
Enable TinyDB class for subclassing

ASMfreaK:
I want to be able to do something like this:
```python
from tinydb import TinyDB
class YamlStorage(...):
    ...

class CachingYamlTDB(TinyDB):
    DEFAULT_STORAGE = CachingMiddleware(YamlStorage)
```
This PR enables TinyDB for such sublassing.

### DIFF
--- a/tinydb/database.py
+++ b/tinydb/database.py
@@ -76,8 +76,8 @@ class TinyDB(object):
                         with ``args`` and ``kwargs``.
         """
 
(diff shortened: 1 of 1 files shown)


msiemens:
That makes sense, thanks for the contribution!

msiemens:
Now released in v3.5.0 🎉

ASMfreaK:
Yay~
Thanks)

PR #151, opened 2017-09-02
MessagePack support

Code-ReaQtor:
Hi @msiemens,

So I was working with my [IPToCC](https://github.com/Code-ReaQtor/IPToCC) module and one of the issues I've encountered is that the database file is huge and I want to reduce it's size.

My solution was to use [MessagePack](http://msgpack.org/index.html). The current serializer/deserializer I used, [msgpack-python](https://github.com/msgpack/msgpack-python), is as fast as ujson but produces smaller file size (~50% smaller).

I've also added some fix to failing tests when using ujson as a parser.

Hope this will be merged to your `master` branch.

Cheers!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -163,6 +163,13 @@ extensions: Contributions to TinyDB are welcome! Here's how to get started:
 Changelog
 *********
 
+**Unreleased** (2017-09-02)
+===========================
+
+- Added MessagePack support.
+- Added skipIf on tests that fails when using ujson.
+
+
 **v3.5.0** (2017-08-30)
 =======================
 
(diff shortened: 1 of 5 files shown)


msiemens:
Hey @Code-ReaQtor, thanks for the contribution! It's great to see extensions for TinyDB being developed. With regards to including this into the TinyDB core, I think the better way would be to create an extension package (something like `tinydb-msgpack`) as in my opinion the TinyDB core should be dependency free. If you set up a repo/PyPI package for your MessagePack storage, I'll gladly link to it in the extension list (https://github.com/msiemens/tinydb#extensions)

Code-ReaQtor:
Hi @msiemens, Sure! Will create an extension then.

PR #158, opened 2017-09-13
Docs: Use the term `documents` instead of `elements`

msiemens:
TinyDB is a document-oriented database. Yet, since the the initial commit (b01cf23013c745c551f6bdffc78d737bf8e5f1f2) we used the term _elements_ instead of _documents_ for the database entries. I think it's actually a kind of weird and awkward term, so this PR would rename all usages of `element` to `document`.

Now, this is mostly a documentation change. But we also have a wrapper class called `Element` and also use `eid` to refer to element/documen IDs. My question: is it worth the hassle to have a new breaking change along with a major version bump (v4.0.0) to rename these usages too? We'd then have a `Document` class and refer to IDs as document IDs/`doc_id`.

@eugene-eeo Do you have any insight on this?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -138,7 +138,7 @@ Extensions
 | **Status:**      *stable*
 | **Description:** ``tinydb-smartcache`` provides a smart query cache for
                    TinyDB. It updates the query cache when
-                   inserting/removing/updating elements so the cache doesn't
+                   inserting/removing/updating documents so the cache doesn't
                    get invalidated. It's useful if you perform lots of queries
                    while the data changes only little.
 
@@ -199,7 +199,7 @@ v3.3.1 (2017-06-27)
 v3.3.0 (2017-06-05)
 ===================
 
-- Allow iterating over a database or table yielding all elements
+- Allow iterating over a database or table yielding all documents
   (see `pull request #139 <https://github.com/msiemens/tinydb/pull/139>`_).
 
 v3.2.3 (2017-04-22)
@@ -217,7 +217,7 @@ v3.2.2 (2017-01-16)
 v3.2.1 (2016-06-29)
 ===================
 
-- Fix a bug with queries on elements that have a ``path`` key
+- Fix a bug with queries on documents that have a ``path`` key
   (see `pull request #107 <https://github.com/msiemens/tinydb/pull/107>`_).
 - Don't write to the database file needlessly when opening the database
   (see `pull request #104 <https://github.com/msiemens/tinydb/pull/104>`_).
@@ -243,7 +243,7 @@ v3.1.3 (2016-02-14)
 v3.1.2 (2016-01-30)
 ===================
 
-- Fix a bug when using unhashable elements (lists, dicts) with
+- Fix a bug when using unhashable documents (lists, dicts) with
   ``Query.any`` or ``Query.all`` queries
   (see `a forum post by karibul <https://forum.m-siemens.de/d/4-error-with-any-and-all-queries>`_).
 
@@ -258,9 +258,9 @@ v3.1.1 (2016-01-23)
 v3.1.0 (2015-12-31)
 ===================
 
-- ``db.update(...)`` and ``db.remove(...)`` now return affected element IDs
+- ``db.update(...)`` and ``db.remove(...)`` now return affected document IDs
   (see `issue #83 <https://github.com/msiemens/tinydb/issues/83>`_).
-- Inserting an invalid element (i.e. not a ``dict``) now raises an error
+- Inserting an invalid document (i.e. not a ``dict``) now raises an error
   instead of corrupting the database (see
   `issue #74 <https://github.com/msiemens/tinydb/issues/74>`_).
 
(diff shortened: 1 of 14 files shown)


dagraham:
I apologize if this is the wrong place for this but I have some example of the serialization extension that I thought might be useful. I'm particularly happy with the DateTimeSerializer. Please let me know if there is a better way to contribute.

`
class DateTimeSerializer(Serializer):
    """
    This class handles both aware and naive datetime objects. 

    Encoding: If the datetime object is aware, it is first converted to UTC and then encoded with an 'A' appended to the serialization. Otherwise it is serialized without conversion and an 'N' is appended.

    Decoding: If the serialization ends with 'A', the datetime object is treated as UTC and then converted to localtime. Otherwise, the datetime object is treated as localtime and no conversion is necessary.

    This serialization discards both seconds and microseconds but preserves hours and minutes.
    """

    OBJ_CLASS = datetime

    def encode(self, obj):
        """
        Serialize naive datetimes objects without conversion but with 'N' for 'Naive' appended. Convert aware datetime objects to UTC and then serialize them with 'A' for 'Aware' appended.
        """
        if obj.tzinfo is None:
            return obj.strftime('%Y%m%dT%H%MN')
        else:
            return obj.astimezone(tzutc()).strftime('%Y%m%dT%H%MA')

    def decode(self, s):
        """
        Return the serialization as a datetime object first converting to localtime if the serializaton ends with 'A'. Serializations that end with 'N' are naive or floating datetimes and are interpreted as already representing localtimes. 
        """
        if s[-1] == 'A':
            return datetime.strptime(s[:-1], '%Y%m%dT%H%M').replace(tzinfo=tzutc()).astimezone(tzlocal())
        else:
            return datetime.strptime(s[:-1], '%Y%m%dT%H%M')


class DateSerializer(Serializer):
    OBJ_CLASS = date  # The class handles date objects

    def encode(self, obj):
        """
        Serialize the naive date object without conversion.
        """
        return obj.strftime('%Y%m%d')

    def decode(self, s):
        """
        Return the serialization as a date object.
        """
        return datetime.strptime(s, '%Y%m%d').date()


class TimeDeltaSerializer(Serializer):
    OBJ_CLASS = timedelta  # The class handles timedelta objects

    def encode(self, obj):
        """
        Serialize the timedelta object as days.seconds.
        """
        return "{0}.{1}".format(obj.days, obj.seconds)

    def decode(self, s):
        """
        Return the serialization as a timedelta object.
        """
        days_seconds = (int(x) for x in s.split('.')) 
        return timedelta(*days_seconds)

serialization = SerializationMiddleware()
serialization.register_serializer(DateTimeSerializer(), 'TinyDateTime')
serialization.register_serializer(DateSerializer(), 'TinyDate')
serialization.register_serializer(TimeDeltaSerializer(), 'TinyTimeDelta')

`

msiemens:
@dagraham That's really cool. I think the best way to discuss this would be the [tinydb-serialization repo](https://github.com/msiemens/tinydb-serialization). Would you mind opening an issue over there?

dagraham:
Absolutely, I’ll be glad to do so.

Thanks,
Dan



On September 19, 2017 at 3:55:26 PM, Markus Siemens (notifications@github.com<mailto:notifications@github.com>) wrote:

@dagraham<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_dagraham&d=DwMCaQ&c=imBPVzF25OnBgGmVOlcsiEgHoG1i6YHLR0Sj_gZ4adc&r=cjSGCFDvj7Zy-plAeyNJjDIBRBeNSNbZaJLNyuBRtxo&m=Wd7FoBY1kUoxZ2FJw1aHRC18fjZ8aRr1dIv6UXE2y9U&s=jTaYCxEepJQVN7n9dIzn9Mbh6pZAwKC9DxHFFMVMYxo&e=> That's really cool. I think the best way to discuss this would be the tinydb-serialization repo<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_msiemens_tinydb-2Dserialization&d=DwMCaQ&c=imBPVzF25OnBgGmVOlcsiEgHoG1i6YHLR0Sj_gZ4adc&r=cjSGCFDvj7Zy-plAeyNJjDIBRBeNSNbZaJLNyuBRtxo&m=Wd7FoBY1kUoxZ2FJw1aHRC18fjZ8aRr1dIv6UXE2y9U&s=jAPIkGmtQ5ocWDo8wv634mGMrwroDKkDJT9oWp1CiiE&e=>. Would you mind opening an issue over there?

eugene-eeo:
Hi, sorry for the very late response – I've been busy lately. I think it is good that we keep it consistent, i.e. go with the `doc_id` and `Document` classes. However this is not a major issue, so when something else comes up that requires a major version bump then we can rename the classes and attributes. At the moment just using properties and warnings would be alright, e.g.:

```
class Document(dict):
    ...

    @property
    def eid(self):
        warnings.warn(...)
        return self.doc_id

Element = Document
```

msiemens:
> At the moment just using properties and warnings would be alright

Good point! Somehow I didn't think of just keeping Element/eid as deprecated and forward them to Document/doc_id. I think that's the way to go  for now.

msiemens:
Thanks @eugene-eeo for your feedback and no worries about the delayed response, often enough I myself need a couple more days to catch up with GitHub issues 🙂

PR #161, opened 2017-10-11
Convert dump data to raw dictionary in YAMLStorage

msembinelli:
Previously, dump was causing the python object to get
dumped instead of the raw dictionary itself.

As a result upon issuing a db.update({'name': 'foo'}) call, the entry would change to:

1: !!python/object/new:tinydb.database.Document
    dictitems: {name: foo}
    state: {doc_id: 1}

Instead of:

1: {name: foo}

Not sure if this is the proper fix, but it seems to work.

Another person described the same issue here:
https://gist.github.com/fndari/bc4d724a52c9d4b1152ea1a2e5a5ffed

### DIFF
--- a/docs/extend.rst
+++ b/docs/extend.rst
@@ -35,7 +35,7 @@ Let's look how you could add a `YAML <http://yaml.org/>`_ storage using
 
         def write(self, data):
             with open(self.filename, 'w') as handle:
-                yaml.dump(data, handle)
+                yaml.dump(yaml.safe_load(str(data)), handle)
 
         def close(self):  # (4)
             pass


msiemens:
Hey @msembinelli,

thanks for the PR! I've thought about this a little. Basically, to store data, this approac first serializes the dict as a string, then loads it using PyYAML and then again stores it using PyYAML. I think, a better and more clean way would be using PyYAML's representer mechanism. I played with it a little and have a commit ready. If it's fine with you, I'd close this PR and push the changes. What do you think?

msembinelli:
I'm all for it. I would rather it be done right than see a hacky solution like mine :). I read a little bit about representers but wasn't sure!

Issue #147, opened 2017-08-12
Add way to get a table's name


msiemens:
See https://forum.m-siemens.de/d/23-table-name-from-the-table-object

My main concern is how to add this in a backwards-compatible way, as I'm not totally sure what guarantees we specify in the docs about what arguments the table class should take.

Phuket2:
@msiemens, wouldn't you just add a readonly attr to store/read the table name?

Phuket2:
@msiemens, I made 3 changes in the database.py to store and access the table name.  I didn't want to do a pull request etc... I am not good enough for that, i am still a beginner.   Also,  its probably not so Pythonic  what i have done.  Anyway, i just wanted to give it a go.  But it does seem like this would not change the interface. Not sure about custom tables etc...
In the def table method, I updated the options kwargs with the name of the table, which is present.  Then in the Table class init, i added a new kwarg param as well as a attr and a property to store and access the attr.  In my simple cases, it appears to work.
 
```python
def table(self, name=DEFAULT_TABLE, **options):
		"""
		Get access to a specific table.
		
		Creates a new table, if it hasn't been created before, otherwise it
		returns the cached :class:`~tinydb.Table` object.
		
		:param name: The name of the table.
		:type name: str
		:param cache_size: How many query results to cache.
		"""
		
		if name in self._table_cache:
			return self._table_cache[name]
			
		options.update({'table_name':name})	# <--- added this
		table = self.table_class(StorageProxy(self._storage, name), **options)
		
		self._table_cache[name] = table
		
		# table._read will create an empty table in the storage, if necessary
		table._read()
		
		return table

class Table(object):
	"""
	Represents a single TinyDB Table.
	"""
	
	def __init__(self, storage, cache_size=10, table_name=''): # added new kwarg
		"""
		Get access to a table.
		
		:param storage: Access to the storage
		:type storage: StorageProxyus
		:param cache_size: Maximum size of query cache.
		"""
		self.__table_name = table_name # <-- private attr added
		self._storage = storage
		self._query_cache = LRUCache(capacity=cache_size)
		
		data = self._read()
		if data:
			self._last_id = max(i for i in data)
		else:
			self._last_id = 0
		
	# property added
	@property
	def name(self):
		return self.__table_name
	
	# class continues...
	
	```

msiemens:
[...]

PR #165, opened 2017-10-27
Sort patch - Initial Pull request to include a sort method

alshapton:
I'd like to have this considered as an additional method to TinyDB. Should it be accepted, I'll amend the documentation to reflect it.


alshapton:
My first Pull request  - it worked locally - failed the Travis CI build - Will fix !!!! Bear with me !

alshapton:
MMMM I dont understand how I can make the code coverage 100% - the line of code that's not covered seems to be a "for" loop - which I know is executed in all use cases and all tests I have designed....... Any ideas from the comunity ?

msiemens:
@alshapton Thanks for the PR! I really appreciate your work, so it's a tough decision to me, wether this should go into the TinyDB core or not. I tend to think that this is better done by the users themselves. Considering the API sorting to me is something that I want to do on the _results_ of a search, but not on a database or table. `db.sort(…)` would implicate to me that I somehow sort the documents inside the database itself. A cleaner way would be something like `db.search(…).sort(…)` which I consider close enough to something like `sorted(db.search(…), key=itemgetter('name'))`.

@eugene-eeo What's your opinion on that?

alshapton:
@msiemens I kinda see your point, but I was happy to get involved  and attempt to add some functionality - will be happy to refactor it into some extension if that's what's required.
I'll wait for you and @eugene-eeo to come back with a direction.....

alshapton:
OK - what should we do - merge this or leave it as a small add-o n useful script listed somewhere ? I dont mind either way !!!

msiemens:
If you want to, you can create a Wiki page called _Useful snippets_ or something like this for this and similar content.

alshapton:
Willdo !

Andrew

eugene-eeo:
For me this is too general to be included in TinyDB since it has very little to do with the underlying database, i.e. querying, IO operations, etc. It is easy to write a key function that one could pass to `sorted` to order data by specific keys. I think a good way to gauge whether we should include something is to ask ourselves: Could I do x better given the knowledge I have? E.g:

 - We can do querying better since:
    - we know the underlying file format
    - we have a query cache

Of course exceptions could be made where necessary for pragmatism, but I hope you see my point.

alshapton:
@eugene-eeo  I do agree - so I think one of my original questions to @msiemens still holds though - is there a roadmap for tinydb and the desirable features that it should contain and/or external libraries and middleware and so on ....?

Also - I'll put something on the wiki page re: this , but how then do I "cancel" the PR ?

msiemens:
> is there a roadmap for tinydb and the desirable features that it should contain and/or external libraries and middleware and so on ....?

That's a good question. As a matter of fact there's no real roadmap for TinyDB. In my mind it's more or less complete since version 3.0. Since the release of 3.0 we've mainly had bugfixes with occasional feature releases for rather small stuff (like specifying the default table name in #98 or accessing a table's name in #147). I'd prefer to have everything else implemented in external libraries/middlewares/extensions and so on. I've outlined some of my thoughts on that in https://blog.m-siemens.de/notes-on-tinydb/.

Regarding extensions, one thing we maybe might want to add is a better extension mechanism. Right now you can extend TinyDB's methods with a custom table class which unfortunately aren't composable. A plugin mechanism might fix that. On the other hand, a plugin mechanism means a lot of complexity which itself may contain or lead to bugs, so I'm undecided on that.

PR #167, opened 2017-10-28
Feature: Upsert

msiemens:
As discussed in https://forum.m-siemens.de/d/30-primary-key-well-sort-of


msiemens:
@alshapton Fixed, thanks!

PR #170, opened 2017-11-26
Fix reference

jean:
.../tinydb/docs/getting-started.rst:34:
WARNING: undefined label: document_ids (if the link has no caption the
label must precede a section header)


msiemens:
Thanks!

PR #171, opened 2017-11-27
Corrected URL and status of TinyMP to be identical to README

alshapton:
The information for TinyMP was incorrect. This PR corrects it.


msiemens:
Oh, sorry for mixing this up and thanks for the fix :)

PR #175, opened 2017-12-12
Use correct symbols in docstrings

cmccandless:
I noticed the symbols were switched in the docstrings in `QueryImpl.__and__` and `QueryImpl.__or__`. This PR corrects this.


msiemens:
Thanks! 🙂

PR #177, opened 2017-12-23
Search inside the list type value as well

davidlatwe:
Example case:
```python
db.insert({
    "user_name": "bob",
    "tag": [
        "Rich",
        "Male",
        "Smart",
    ]
})
db.insert({
    "user_name": "alice",
    "tag": "Smart",
})

table = db.table('_default')
User = Query()
result = table.search(User.tag == "Smart")
print(len(result))

# result: 1
```

Currently it simply compare with value, did not consider value is a list or not.
In the other hand, MongoDB does that, maybe TinyDB could support that, too?

Made a change in `Query` object to achieve that,
and the result from example above should return `2`

### DIFF
--- a/tinydb/queries.py
+++ b/tinydb/queries.py
@@ -147,7 +147,7 @@ class Query(object):
         """
         if sys.version_info <= (3, 0):  # pragma: no cover
             # Special UTF-8 handling on Python 2
-            def test(value):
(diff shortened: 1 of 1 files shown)


davidlatwe:
Sorry, should not be like this, I need to re-think about it.

Issue #185, opened 2018-01-22
Python 2.6 failed to setup in CI test


davidlatwe:
### Problem

In the time I made PR #184 , Python 2.6 start failing to pass the CI test setup, due to `pytest-runner` not able to find the package `setuptools_scm`. Checkout branch `master` to build test also showed same result.

### Discussion

Based on this [fact](https://github.com/msiemens/tinydb/pull/184#issuecomment-359277834), can Python 2.6 be dropped ?

msiemens:
I think this should be fine as pip also dropped support for Python 2.6 (see pypa/pip#3955). If no one else has any objections, we can do this.

@eugene-eeo

hugovk:
Please see PR https://github.com/msiemens/tinydb/pull/188.

PR #187, opened 2018-01-27
Fix 0 capacity issue. now cache_size can be 0

wuarmin:
I encountered a problem while setting cache_size to 0. The result was an unlimited cache because in python "0" evaluates to false. This PR would fix this.

kind regards

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -45,6 +45,17 @@ def test_caching(db):
 
     assert table1 is table2
 
+def test_zero_cache_size(db):
+    table = db.table('table3', cache_size=0)
+    query = where('int') == 1
+
+    table.insert({'int': 1})
+    table.insert({'int': 1})
+
+    assert table.count(query) == 2
+    assert table.count(where('int') == 2) == 0
+    assert len(table._query_cache) == 0
+
 
 def test_query_cache_size(db):
     table = db.table('table3', cache_size=1)
--- a/tinydb/utils.py
+++ b/tinydb/utils.py
@@ -21,7 +21,10 @@ class LRUCache(dict):
                          or ``None`` for an unlimited cache size
         """
 
-        self.capacity = kwargs.pop('capacity', None) or float('nan')
+        self.capacity = kwargs.pop('capacity', None)
+        if self.capacity is None:
+            self.capacity = float('nan')
+            
         self.lru = []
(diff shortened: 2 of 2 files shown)


wuarmin:
@msiemens Do you know why the test-build with Python 2.6 failed?

msiemens:
Thanks for the Pull Request! 🙂

> @msiemens Do you know why the test-build with Python 2.6 failed?

This is due to #185 where it's still up to debate whether or not to drop Python 2.6 as our test infrastructure has dropped support for it.

wuarmin:
@msiemens All right. I 'll provide another to provide the option to disable cache explicitly.

msiemens:
Actually, I'm just thinking about if the best way to disable the query cache would be to actually use a `cache_size` of 0 and in this case skip the query cache logic or replace the cache with an class that has the same interface/api but doesn't perform any caching. This may make the API a little less nosiy compared to adding a new argument – if we document this properly. What do you think?

wuarmin:
@msiemens Yes, I think you are right. Defining a cache_size of zero, fullfills the purpose. Another option is not necessary.

msiemens:
This is now released in v3.8.0 🙂

wuarmin:
@msiemens The latest version in python package index is still 3.7.0.
If I want to upgrade I get following message:

```txt
CRITICAL:pip.index:Could not find a version that satisfies the requirement tinydb==3.8.0 (from versions: 1.0.0, 1.0.1, 1.1.0, 1.1.1, 1.2.0, 1.3.0, 1.4.0, 2.0.0, 2.0.1, 2.1.0, 2.2.0, 2.2.1, 2.2.2, 2.3.0b0, 2.3.1, 2.3.1.post1, 2.3.1.post2, 2.3.2, 2.4, 3.0.0, 3.1.0, 3.1.1, 3.1.2, 3.1.3, 3.2.0, 3.2.1, 3.2.2, 3.2.3, 3.3.0, 3.3.1, 3.4.0, 3.4.1, 3.5.0, 3.6.0, 3.7.0)
```

msiemens:
Whoopsi! Forgot to publish it on PyPI, sorry! Should work now 🙂

wuarmin:
thanks! :+1: I'm on 3.8.0 :smile:

PR #195, opened 2018-03-21
Avoid installing tests as a package.

### LINKED ISSUE #194: Conflict of modules when installing TinyDB
When installing _tinydb_ 3.8.0 the following structure is created:

```
.venv\lib\site-packages\tests\
.venv\lib\site-packages\tinydb-3.8.0.dist-info\
.venv\lib\site-packages\tinydb\
```

The problem is that I also have a package called _tests_. When I try to import any package or module from _tests_ I get the following message:

```
ImportError: no module name xxx
```

I can change my package's name, but I think another user can face the same problem. It was a bit time-consuming.

Is it necessary to have the _tests_ module on the same level as the _tinydb_ module? If it is going to stay on the same level, couldn't be called tinydb_tests or similar? Should tests be part of the package when installed?

tarcisioe:
Fixes #194

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ exec(read('tinydb/version.py'))
 setup(
     name="tinydb",
     version=__version__,
-    packages=find_packages(),
+    packages=find_packages(exclude=['tests']),
 
     # development metadata
     zip_safe=True,


msiemens:
Thanks for the pull request! 🙂

msiemens:
I'll try to get to releasing a new version of TinyDB next week

msiemens:
Released in v3.8.1 now ✨

tarcisioe:
The wheel distributed on PyPI still has the `tests` module. I tried running `python setup.py bdist_wheel --universal` on my own clone of the repository and it was created correctly. Maybe it was generated against an unclean directory?

msiemens:
Sorry, my bad! Yes, I didn't clean the `build/` directory. Now v3.8.1.post1 is released which actually fixes the problem 🙂 Just upgrading to the lastest verison of TinyDB shold install it

arunpersaud:
Is there a way to still keep the tests files in the tar-ball (they are missing in the 3.8.1post1)? Would be nice to be able to run them before doing setup.py install (they shouldn't get installed though I guess).

msiemens:
> Is there a way to still keep the tests files in the tar-ball (they are missing in the 3.8.1post1)? Would be nice to be able to run them before doing setup.py install (they shouldn't get installed though I guess).

I'd be open to that. In my other Python open source project, [PyGitUp](https://github.com/msiemens/PyGitUp), tests are included inside the package (in `PyGitUp/tests` after installation). We could opt for something like this too. @eugene-eeo Do you have an opinion on that?

PR #202, opened 2018-04-05
Fix bug in write_back

eugene-eeo:
- Fixes a bug in `write_back(items)` – it will no longer raise an error if `items` is empty.
 - Make error detection `O(n)` – no longer sort the `doc_ids` array only to retrieve the last element, just take the maximum element instead.


msiemens:
Thanks! 🙂

PR #212, opened 2018-04-23
fsync added following flush

parkerduckworth:
A call to `os.fsync()` will ensure all internal buffers associated with self._handle are written to disk.


msiemens:
Thanks 🙂

Issue #86, opened 2015-12-31
Query using partial keywords


konsbn:
Suppose I have a database of books that stores titles, authors etc. Now If I have a book titles "Complex Analysis" I wish the query using "complex" should show up all the matches in the database in this case "COmplex Analysis". I know perhaps this is not the place to ask this but please help me.

msiemens:
You can search using regular expressions:

    >>> db.search(User.name.matches('[aZ]*'))
    >>> db.search(User.name.search('b+'))

These correspond to Python's [`re.match`](https://docs.python.org/2/library/re.html#re.match) and [`re.search`](https://docs.python.org/2/library/re.html#re.search) where the latter searches for an occurenc that may start anywhere in the string.

Alternatively you can use a custom test expression like so:

    >>> test_contains = lambda value, search: s.contains(search)
    >>> db.search(User.name.test(test_contains, 'Columbia'))

For more details, refer to the [relevant section of the TinyDB docs](http://tinydb.readthedocs.org/en/latest/usage.html#advanced-queries).

BTW: there was a very similar question [on StackOverflow](http://stackoverflow.com/q/34476895/997063).

msiemens:
@konsbn Does my answer solve your question?

msiemens:
Closing due to inactivity.

PR #216, opened 2018-07-13
Add support for regex flags

choyj:
This addresses Issue #86 allowing the user to perform case-insensitive
matches and searches


msiemens:
Thanks! 🙂

msiemens:
I've now released v3.10.0 which includes this feature 🙂

Issue #221, opened 2018-07-28
[FR] use doc_id as id if the dict has doc_id


Cologler:
for example, the id of `db.insert({ 'doc_id': 'abc' })` should be 'abc':

``` json
{
    "_default": {
        "abc": {
            ......
        }
    }
}
```

msiemens:
I don't think we want to allow this directly. That way users can overwrite arbitrary entries while expecting to insert new ones. Also, this would break any usage where someone already uses the `doc_id` field in their database.

msiemens:
@Cologler What's your exact use case for this?

Cologler:
Proposal: insert or replace exists item without enumerate all items

Cologler:
It would not break anything if doc_id in both of key and entity:

``` 
"the Id": { "doc_id": "the Id"  }
```

Cologler:
since tinydb already has a `upsert()` method:

when user call `db.insert({ 'doc_id': 'abc' })`, if the id already exists, raise a `KeyError`.

eugene-eeo:
An alternative would be to have a special `_meta` attribute on `Query()`. (Inspired by Elasticsearch) For instance:

```python
user = Query()
db.get(user._meta.doc_id >= 2)
```

but this would require changing how the Query object works internally. I.e. it would have to receive now two parameters, the document currently being tested and the metadata associated with the document. In `Query._test_generate` we would have to change it to something like:

```python
...
        def impl(value, meta):
            try:
                # Resolve the path
                for part in self._path:
                    value = value[part]
            except (KeyError, TypeError):
                return False
            else:
                return test(value, meta)
...
```

To differentiate between the new `Query()._meta` (query on the _metadata_) and an actual query on the `_meta` field, we would use `Query()['meta']` to mean the latter.

PR #224, opened 2018-08-04
Feat: support user storage proxy for string type doc_id

Cologler:
for #221 .

examples see the added unittest.

### DIFF
--- a/tinydb/database.py
+++ b/tinydb/database.py
@@ -77,7 +77,7 @@ class StorageProxy(object):
         self._storage = storage
         self._table_name = table_name
 
-    def construct_doc(self, key, val):
+    def _new_document(self, key, val):
         doc_id = int(key)
         return Document(val, doc_id)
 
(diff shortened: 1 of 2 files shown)


tinydb/database.py
```diff
@@ -77,6 +77,10 @@ def __init__(self, storage, table_name):
         self._storage = storage
         self._table_name = table_name
 
+    def construct_doc(self, key, val):
```
msiemens: Could we rename this to an internal method and also use something like `def _new_document(…)`?

Cologler: a new commit for that.

msiemens: Awesome! Now you just need to adapt the example in https://github.com/msiemens/tinydb/blob/a454998e65e43f7af44e008129d08a44fcfed1e6/tests/test_tinydb.py#L617-L618 🙂

Cologler: 😂

msiemens:
Thanks for the idea, the contribution and the modifications, @Cologler! I'll try to bundle up a release of TinyDB next week 🙂

PR #227, opened 2018-08-21
allow arguments to be passed to json.load as well as json.dumps

ed588:
this may not be a very elegant way but at least it's backwards-compatable


ed588:
it looks like the dict merging doesn't work in older versions of python, but you could fix that easily enough

Issue #226, opened 2018-08-19
Implement `__repr__` methods for classes `TinyDb`, `Table`, `Query` etc.


cool-RR:

msiemens:
@cool-RR Any suggestions what `__repr__` should contain for these classes? Do you know what other databases do here with Python?

cool-RR:
Table should have the table name, maybe the number of entries. Tinydb
should have the number of tables. Query should have the query.

msiemens:
@cool-RR Would you be willing to write a Pull Request for this?

cool-RR:
No, sorry. I understand if this ticket will be closed.

msiemens:
Okay, I've added the `contributions-welcome` tag and am happy to accept pull request if anyone wants to implement this 🙂

piraces:
Hello! I would like to participate and try implement it. So I will work on it if you’re open 🙂

mrfunnyshoes:
Hello, is this still open? I might wanna contribute...

cool-RR:
I believe it is, go for it Edgar!

Xarvalus:
Hello 🙂

I have opened [WIP] PR on your issue, sorry for getting in your way @mrfunnyshoes with my contribution, I was investing some time for this case since a few days before.

If you are willing to help with the matter feel free to contribute within https://github.com/msiemens/tinydb/pull/229, definiently there will be more work to do 😄

msiemens:
Well, now we have two PRs for this 🙈 @mrfunnyshoes Would you be interested in doing a code review for the patch by @Xarvalus (on #229)?

msiemens:
And I really appreciate both contributions and don't want anyone to feel like their work is not valued, even if it's a 'small' thing like `__repr__`! The PR by @Xarvalus is more exhaustive and covers testing too, that's why inclined towards this one.

PR #229, opened 2018-09-08
__repr__ methods for classes TinyDB, Table and Query

Xarvalus:
Hello, considers https://github.com/msiemens/tinydb/issues/226.

I have added basic `__repr__` implementation for mentioned classes, would like to know if you see it as sufficient and which else classes should be implemented within the same way.

You may try the reprs with:
```
from tinydb import database, queries

print(repr(database.TinyDB('test.json')))
print(repr(database.Table(database.JSONStorage('test.json'), 'some name')))

Fruit = queries.Query()
print(repr(Fruit.type == 'peach'))
```

Achieving results:
```
TinyDB(tables={'_default'}, tables_count=1, default_table_documents_count=0, all_tables_documents_count=['_default=0'])
Table(name='some name', total=1, storage=<tinydb.storages.JSONStorage object at 0x10212c2e8>)
QueryImpl('==', ('type',), 'peach')
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ pip-log.txt
 .coverage
 .tox
 nosetests.xml
+.pytest_cache/
 
 # Translations
 *.mo
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -329,3 +329,10 @@ def test_orm_usage():
     query2 = User.age.year == 2000
     assert query1(data)
     assert query2(data)
+
+
+def test_repr():
+    Fruit = Query()
+
+    assert repr(Fruit) == "Query()"
+    assert repr(Fruit.type == 'peach') == "QueryImpl('==', ('type',), 'peach')"
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -112,5 +112,7 @@ def test_table_repr(db):
     name = 'table4'
     table = db.table(name)
 
-    assert bool(re.match(r"\<Table name\=\'table4\', total\=0, storage=\<tinydb\.database\.StorageProxy object at [a-z0-9]+\>\>", repr(table))) is True
-
+    assert re.match(
+        r"<Table name=\'table4\', total=0, "
+        "storage=<tinydb\.database\.StorageProxy object at [a-z0-9]+>>",
+        repr(table))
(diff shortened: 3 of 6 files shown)


cool-RR:
Thanks for volunteering to help.

The first two reprs are wrong because they use a Foo(bar) format which should only be used if the object could be reconstructed completely from Foo(Bar). I think Query can but the others can't.
So you need something like `<Table name='foo' n_entries=7>`

cool-RR:
Also I'm not sure that you ran your code. I suggest adding tests for these `__repr__` methods.

Xarvalus:
Thanks for feedback.

I didn't add tests as `__repr__` method are stripped away from test coverage, the above example is fairly easy and minimal to just one-time run with temporary script file for manual testing but if we prefer to spread more repr's across the classes I agree and will add those to tests.

I will adjust to the proper format shortly.

What about the sufficient info in `__repr__`s? Anything more I should add? Any thumb rule about it?

Also which classes do you prefer to also implement reprs?

cool-RR:
1. I think that tests would be good.
2. Looking forward to seeing your new commit.
3. You've got sufficient information there.
4. These three classes are great for now.

Xarvalus:
@cool-RR I have updated the PR, please kindly take a look

On 2.7 one of test failed, caused by https://stackoverflow.com/questions/9773121/removing-u-in-list do you suggest any best practice for keeping result the same across 2.x/3.x?

cool-RR:
Looks good to me. I'm not the project owner, so he'll need to approve before merging.

cool-RR:
I don't think it's important to keep the result identical in Python 2 and 3.

Xarvalus:
Hey @msiemens meanwhile I have changed TinyDB's class repr test into regex one, just to match the Python 2.7's `u` strings issue.

Currently it passes all the checks, let know if you would like any adjustments

Xarvalus:
Oh, sorry.. I didn't see you wrote about 2 mins before me 😄 @cool-RR okay, will push your suggestions

cool-RR:
Great. Looks good to me. I'm not the project owner, so he'll need to approve before merging.

msiemens:
Thanks for your work  @Xarvalus, it's much appreciated 🙂

msiemens:
This is now released in TinyDB v3.11.1 🙃

cool-RR:
Awesome! Thanks for the quick release!

Issue #222, opened 2018-07-28
[FR] add encoding argument for JSONStorage


Cologler:
Currently, JSONStorage use `open(..., encoding=None)` to open the json file. 

But I try to use utf-8 encoding.

msiemens:
Hey @Cologler, could you provide an example what you write to the database, what the database file looks like and which OS and Python version you use?

As far as I can tell, Python uses the default encoding (as determined by `locale.getpreferredencoding()`) which in many cases is UTF 8 (but Windows for example may differ). Now, non-ASCII characters will be escaped unless passing `ensure_ascii=False` to the TinyDB constructor:

```python
db_file = tmpdir.join('test.db')
db = TinyDB(str(db_file), ensure_ascii=False)

# Write contents
db.insert({'a': 'ברי צקלה'})

assert db.all() == [{'a': 'ברי צקלה'}]
assert db_file.read() == '''{"_default": {"1": {"a": "ברי צקלה"}}}'''
```

Cologler:
In my windows, `locale.getpreferredencoding()` return `cp936`.

I use some editor to edit the json file (save to utf-8), and tinydb cannot open it agains.

msiemens:
Okay, fair enough, that I can try to reproduce somehow :)

msiemens:
@Cologler Can you post the exact error that occurs while reading a UTF-8 database file modified from an external editor? I'm still trying to come up with a test case for the test suite and it always works without any modifications…

msiemens:
As far as I can tell, Python's JSON module always uses UTF-8 by default, so I don't know why this wouldn't work

Cologler:
The external json editor is not ensure asscii, so you can simple to add test case:

``` py
db_file = 'test.db'
with tinydb.TinyDB(db_file) as db:
    db.insert({'a': 'ברי צקלה'})

with open(db_file, 'r', encoding='utf-8') as fp:
    data = json.loads(fp.read())

with open(db_file, 'w', encoding='utf-8') as fp:
    fp.write(json.dumps(data, ensure_ascii=False))

# should not raise:
tinydb.TinyDB(db_file)
```

msiemens:
@Cologler Funny, your example already works for me, both on Python 2 and 3…

msiemens:
> @msiemens why does the tag v3.11 in github repo not match v3.11 on pypi?

I just forgot to actually change the version number when tagging the release. I then adjusted the version number and actually uploaded the release to PyPI.

[...]

PR #238, opened 2018-10-01
Add encoding option + fixed test that fails on windows.

Zwork101:
To solve one of the issues presented in #222, the very simple option was allow the encodings kwarg to be passed in. I also added a test for it. After running the tests, I noticed `test_table_repr` failed. I believe this is because on linux based machines, it displays the memory location with lowercase letters, while windows uses uppercase and lowercase.

### DIFF
--- a/tinydb/storages.py
+++ b/tinydb/storages.py
@@ -75,7 +75,7 @@ class JSONStorage(Storage):
     Store the data in a JSON file.
     """
 
-    def __init__(self, path, create_dirs=False, encoding: str=None, **kwargs):
+    def __init__(self, path, create_dirs=False, encoding=None, **kwargs):
         """
         Create a new instance.
 
--- a/tinydb/storages.py
+++ b/tinydb/storages.py
@@ -4,9 +4,8 @@ implementations.
 """
 
 from abc import ABCMeta, abstractmethod
-import io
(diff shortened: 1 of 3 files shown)


Zwork101:
I'm going to work on this some more, I just need to use io.open instead. I'm not great at this cross-compatibility stuff.

tinydb/storages.py
```diff
@@ -88,7 +90,7 @@ def __init__(self, path, create_dirs=False, **kwargs):
         super(JSONStorage, self).__init__()
         touch(path, create_dirs=create_dirs)  # Create file if not exists
         self.kwargs = kwargs
-        self._handle = open(path, 'r+')
+        self._handle = io.open(path, 'r+', encoding=encoding)
```
msiemens: I think if you use [`codecs.open`](https://docs.python.org/2/library/codecs.html#codecs.open), you don't have to use `serialized = unicode(serialized)` below, which would make it cleaner in my eyes. Could you test if that works?

msiemens:
@Zwork101 Could you have a look at the code review?

msiemens:
No worries! Thanks for your contribution ❤️

msiemens:
This is now released in v3.12.0 🎉

Issue #240, opened 2018-10-05
me thinks typo in


kevindawson:
me thinks typo in:

 * tinydb/database.py

```python
def _get_doc_id(doc_id, eid):
    # Backwards-compatibility shim
    if eid is not None:
        if doc_id is not None:
            raise TypeError('cannot pass both eid and doc_id')

        warnings.warn('eid has been renamed to doc_ids', DeprecationWarning)
        return eid
    else:
return doc_id
```

should be:
```warnings.warn('eid has been renamed to doc_id', DeprecationWarning)```

ps. found testing whilst testing with 3.7
regards

msiemens:
Closed in #241 🙃

PR #241, opened 2018-10-08
Fix typo in warning, should be singular

dastagg:


dastagg:
This would fix https://github.com/msiemens/tinydb/issues/240

msiemens:
Thanks! 🙂

Issue #235, opened 2018-09-25
JSONStorage __init__() modifies database Modify and Change times


sserrata:
I noticed that `touch(path, create_dirs=create_dirs)` is called during `JSONStorage` initialization which executes `os.utime(fname, None)` on the database file. I'm looking for an explanation as to why this is necessary as it appears to be part of the initial commit?

I wouldn't expect the database file `modify` or `change` times to update unless a database operation is performed that actually modifies the file or the file permissions are altered.

msiemens:
Hey @sserrata, this behavior has been there since the initial commit in b01cf23013c745c551f6bdffc78d737bf8e5f1f2 and no one really bothered to change it so far 🙃 Would you be willing to help out with a pull request?

sserrata:
Hi @msiemens - sure, I'd be honored to submit a PR :smile:.

msiemens:
Closing this as it has been fixed in #242 🙃

PR #242, opened 2018-10-11
Check if storage file exists before creating it to avoid needlessly altering mtime

sserrata:
Resolves an issue described in #235 where `JSONStorage` was altering the storage file modify time each time it was initialized. This PR introduces a preliminary check to see if the storage file exists before attempting to create it.


msiemens:
Thanks! 🙂

Issue #218, opened 2018-07-13
passing a single dict to insert_multiple corrupts the tinydb file


Phuket2:
I did something like the below by mistake (instead of passing a list of dicts, I passed a single dict).  TinyDB does not complain and appears to write a malformed entry to the file which prevents the file from being reopened thereafter.

I tried this on Python 3.6 and 3.7 and got the same results.  I am using tinydb version -  3.9.0.post1

```python
from tinydb import TinyDB

db = TinyDB("badtest.json")
d = dict(first='John', last='smith')
db.insert_multiple(d)
db.close()
```

msiemens:
That's an interesting design question. On one hand, it's clearly using the API in an erronous way. On the other hand, it's an easy mistake to make. We could add a type check here (i.e. `isinstance`), but that would incur a small performance penalty on everyone else, who is using the API correctly. Right now, I'm undecided about how to approach and solve this issue best…

user-na:
Just thinking out loud, maybe add the check and give a possibility to skip the check with an optional argument (e.g. unchecked=True) for those who need the speed?

msiemens:
> Just thinking out loud, maybe add the check and give a possibility to skip the check with an optional argument (e.g. unchecked=True) for those who need the speed?

In my opinion that would be more confusing than helpful as it's really unexpected and not consistent with the rest of the API interface. Effectively, whether we like it or not, right now we have the _garbage in, garbage out_ approach. Although I'd be interested in how other Python projects approach problems like this one.

circa10a:
@msiemens
What if `insert_mulitple` was deprecated or simply rerouted to `insert` which performed the `isinstance` check and automatically created an single instance array if `isinstance` returned false.
Yes, there would be a slight performance hit, but IMO will eventually simplify the API to have 1 insert method and would be less error prone.

msiemens:
> What if insert_multiple was deprecated or simply merged with insert which performed the isinstance check and automatically created a single instance array if isinstance returned false.

I think keeping `insert` and `insert_multiple` separate is a better and more consistent API. I really prefer having more single purpose methods than fewer multi purpose ones. In the end it's a trade-off between having too many methods, which is confusing, or having too few methods which do everything, which is confusing too.

Mahgoobi:
@msiemens how much is the performance penalty of such addition?

msiemens:
[...]

PR #245, opened 2018-10-30
Internal: refactor isinstance-dict to isinstance-MutableMapping

caichinger:


msiemens:
Thanks for the PR, @caichinger! One small request: could you add a small test case, for example in `tests/test_tinydb.py`?

caichinger:
@msiemens sure, I should have thought of that myself :)

I added [two small tests](https://github.com/msiemens/tinydb/pull/245/commits/ef01ac1727f31141a12065a365bc2ad93cd3cbd7) to `tests/test_tinydb.py` as you suggested.

However, it occurred to me that checking for `Mapping` only is actually sufficient and the current version is therefore based on `Mapping`.

In my opinion in favor of `Mapping`.
- less strict
- sufficient

In favor of `MutableMapping`:
- more in line with the dict-like (mutable) nature of the documents

Personally, I am inclined towards `Mapping` since the check only occurs upon insertion and at this point a `Mapping` (mutable or not) is needed.

What would you prefer? I would then make the necessary changes.

eugene-eeo:
@caichinger just my 20c - `Mapping` better describes what we want from the object (i.e. that we only want to read from it). Since we do not modify the document when we generate an ID for it, this is fine. 

From a backwards compatibility POV, `MutableMapping` might be better (since it is a relaxation of `isinstance(x, dict)`) since we can always relax it more or decide not to, whereas if we decide that we want `isinstance(x, MutableMapping)` in the future then we'd have to bump a major API version.

msiemens:
I think I'm totally fine with using `Mapping`. Semantically it seems like the right choice as we don't really want to modify documents while they are being stored in the database.

caichinger:
@eugene-eeo @msiemens thank you two for your comments.

Please take a look at the PR and let me know if you want me to make any amendments.

msiemens:
@caichinger  Thanks for your contribution! 🙂

msiemens:
This is now released in v3.12.0 🎉

caichinger:
@msiemens cool, thank you for accepting my suggestion! :)

tests/test_tinydb.py
```diff
@@ -92,6 +92,29 @@ def test_insert_multiple_with_ids(db):
...
+            self.data = data
+        def __getitem__(self, key):
+            return self.data[key]
+        def __iter__(self):
+            return iter(self.data)
+        def __len__(self):
+            return len(self.data)
+
+    db.purge()
+    db.insert(CustomDocument({'int': 1, 'char': 'a'}))
```
halfdan: This test is actually failing when JSONStorage is used.

caichinger: That is correct, when using JSONStorage, a custom encoder needs to be passed.

Something like below would work.

````python
class CustomDocument(Mapping):
    def __init__(self, data):
        self.data = data
    def __repr__(self):
        return f'{self.__class__.__name__}({self.data})'
    def __getitem__(self, key):
        return self.data[key]
    def __iter__(self):
        return iter(self.data)
    def __len__(self):
        return len(self.data)


class CustomEncoder(JSONEncoder):
    def default(self, obj):
        return obj.data

db = TinyDB('path/to/to.json', cls=CustomEncoder)
db.insert(CustomDocument({'int': 1, 'char': 'a'}))
assert db.count(where('int') == 1) == 1
````

(The `db` fixture (obviously) uses MemoryStorage.)

What is your suggestion?

Comment (msiemens) [reply tests/test_tinydb.py]: Oh, thanks for reporting this, @halfdan! It's really unlucky that Python's `json` module doesn't support `collections.abc.Mapping`. There's a [question on StackOverflow](https://stackoverflow.com/q/30062760/997063) regarding the rationale behind this behavior, but no clear answer is given.

As a workaround we could convert all documents to `dict`s during insert using `dict(mapping)` like this:

```python
data[doc_id] = dict(document)
```

We'd have to do that for inserts as well as every other operation that is adding/updating documents to the database.

Coincidentally this would also fix #218.

msiemens: I think this would be the best way to solve it. I've created a PR at #256 to discuss this idea.

caichinger: I too believe that fixing this behavior internally makes sense because not providing a custom converter is an easy mistake to make. I added a few comments to #256.

PR #247, opened 2018-11-06
Use tuple literals instead of tuple class/constructor

vstoykov:
This can be a micro optimization (blind guess I was not done any measurements) because `tuple` name will not be looked in the global scope and also there will not be extra function call.

### DIFF
--- a/tinydb/queries.py
+++ b/tinydb/queries.py
@@ -105,10 +105,10 @@ class Query(QueryImpl):
     """
 
     def __init__(self):
-        self._path = []
+        self._path = ()
         super(Query, self).__init__(
             self._prepare_test(lambda _: True),
-            ('path', tuple(self._path))
+            ('path', self._path)
         )
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


msiemens:
It's not a big difference, but I'll take it 🙃 Thanks for the contribution!

```python
In [23]: %timeit QueryOld()['value']['val']
8.91 µs ± 23.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [22]: %timeit Query()['value']['val']
7.47 µs ± 21.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [25]: %timeit QueryOld().value < 1
8.83 µs ± 63.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [24]: %timeit Query().value < 1
7.69 µs ± 53.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

eugene-eeo:
Side note: I think what I thought back then was that the attribute chaining (e.g. `Query().a.b`) should be fast and somehow I used a list instead. But you're right, this is much better since it avoids creating new lists and converting them into tuples. :sweat_smile:

msiemens:
FYI, this is is now included in TinyDB v3.12.2 🎉

Issue #248, opened 2018-11-07
search multi times throw exception


TheBigFish:
```python
db = TinyDB('abc.json', sort_keys=True, indent=2)
db.purge()
db.insert({'type': 'apple', 'count': 7})
db.insert({'type': 'peach', 'count': 3})
db.search(where('count') == 2)
db.search(where('count') == 2)
```
I get this:
```bash
[]
Traceback (most recent call last):
  File "/home/turtlebot/robot_ws/src/simple_server/scripts/test_tinydb.py", line 232, in <module>
    print db.search(where('count') == 2)
  File "/home/turtlebot/.local/lib/python2.7/site-packages/tinydb/database.py", line 593, in search
    return self._query_cache[cond][:]
TypeError: 'NoneType' object has no attribute '__getitem__'
```

joamag:
I'm also getting this error since version 3.12.0

AndrewDRX:
I have created pull request #249 for this issue.

marinang:
I also have the same problem!

msiemens:
Thanks to the PR by @AndrewDRX this is now fixed in v3.12.1. Thanks @AndrewDRX!

PR #249, opened 2018-11-08
Fix for issue #248

AndrewDRX:
Fix for issue #248


msiemens:
Thanks for the fix! ❤️

PR #255, opened 2018-11-17
Add aiotinydb in extensions

ASMfreaK:
I've written a simple shim to make TinyDB more async-aware.


msiemens:
Thanks for adding it to the extension list 🙂

PR #256, opened 2018-11-17
Convert documents to dicts during insertion

### LINKED ISSUE #218: passing a single dict to insert_multiple corrupts the tinydb file

msiemens:
See discussion in https://github.com/msiemens/tinydb/pull/245/files#r231980817 for background

Coincidentally closes #218

@halfdan @caichinger @eugene-eeo

### DIFF
--- a/tests/test_tinydb.py
+++ b/tests/test_tinydb.py
@@ -103,16 +103,57 @@ def test_insert_valid_mapping_type(db):
     class CustomDocument(Mapping):
         def __init__(self, data):
             self.data = data
+
         def __getitem__(self, key):
             return self.data[key]
+
         def __iter__(self):
             return iter(self.data)
+
         def __len__(self):
             return len(self.data)
 
     db.purge()
     db.insert(CustomDocument({'int': 1, 'char': 'a'}))
     assert db.count(where('int') == 1) == 1
+
+
+def test_cutom_mapping_type_with_json(tmpdir):
+    from tinydb.database import Mapping
+
+    class CustomDocument(Mapping):
+        def __init__(self, data):
+            self.data = data
+
+        def __getitem__(self, key):
+            return self.data[key]
+
+        def __iter__(self):
+            return iter(self.data)
+
+        def __len__(self):
+            return len(self.data)
+
+    # Insert
+    db = TinyDB(str(tmpdir.join('test.db')))
+    db.purge()
+    db.insert(CustomDocument({'int': 1, 'char': 'a'}))
+    assert db.count(where('int') == 1) == 1
+
+    # Insert multiple
+    db.insert_multiple([
+        CustomDocument({'int': 2, 'char': 'a'}),
+        CustomDocument({'int': 3, 'char': 'a'})
+    ])
+    assert db.count(where('int') == 1) == 1
+    assert db.count(where('int') == 2) == 1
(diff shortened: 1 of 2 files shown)


caichinger:
I have not yet taken a look at the changes but since converting to `dict` or performing `isinstance` checks incurs an overhead (and was discussed in #218 ), I thought I provide a few primitive timings using [IPython's %timeit](https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-timeit)

````python
In [1]: from collections.abc import Mapping
   ...: 
   ...: class CustomDocument(Mapping):
   ...:     def __init__(self, data):
   ...:         self.data = data
   ...:     def __repr__(self):
   ...:         return f'{self.__class__.__name__}({self.data})'
   ...:     def __getitem__(self, key):
   ...:         return self.data[key]
   ...:     def __iter__(self):
   ...:         return iter(self.data)
   ...:     def __len__(self):
   ...:         return len(self.data)
   ...: 
   ...: 
   ...: minimal_dict = {'int': 1}
   ...: minimal_doc = CustomDocument(minimal_dict)
   ...: another_dict = {key: list(range(1_000)) for key in range(10)}
   ...: another_doc = CustomDocument(another_dict)

In [2]: %timeit "{'int': 1}"
7.78 ns ± 0.154 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)

In [3]: %timeit isinstance(minimal_dict, dict)  # of course similar for another_dict
66.1 ns ± 1.16 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %timeit dict(minimal_dict)
181 ns ± 4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit dict(minimal_doc)
1.63 µs ± 7.32 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [6]: %timeit dict(another_dict)
360 ns ± 3.37 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [7]: %timeit dict(another_doc)
3.24 µs ± 23 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
````

To me it seems that converting a dict into a dict is quite fast - and this is the most likely scenario.
Changing the API to avoid a check/conversion (suggested in #218) is not really a good idea in my opinion; converting the Mapping takes longer but I too think that and not providing a custom converter is a too easy mistake to make (as addressed in #245 ).

From my point of view the performance overhead seems okay given that the library's selling point is simplicity and ease of use, not performance. However, I acknowledge that I do not know how it is used by others.


If desired/required, I can perform additional benchmarks.

msiemens:
Thanks @caichinger for the microbenchmarks! I agree with your conclusion. Using `dict(document)` simply looks like the most clean (i.e. pythonic) solution. We'll have to convert `Mapping`s to `dict`s anyway so I think this is the way to go.

PR #259, opened 2019-02-02
Feature (storages): Add Ephemeral JSON Storage.

imalento:
Hi! I have made tinydb work on GoogleAppEngine.
Please try using it.

- Open JSON read only
- Write on memory

### DIFF
--- a/tinydb/storages.py
+++ b/tinydb/storages.py
@@ -134,3 +134,44 @@ class MemoryStorage(Storage):
 
     def write(self, data):
         self.memory = data
+
+
+class EphemeralJSONStorage(Storage):
+    """
+    Use JSON file readonly.
+    """
+
+    def __init__(self, path, encoding=None, **kwargs):
+        """
+        Create a new instance.
+
+        :param path: Where to store the JSON data.
+        :type path: str
+        """
+
+        super(EphemeralJSONStorage, self).__init__()
+        self.kwargs = kwargs
+        self._handle = codecs.open(path, 'r', encoding=encoding)
+        self.memory = None
+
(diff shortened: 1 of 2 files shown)


tinydb/storages.py
```diff
@@ -134,3 +134,44 @@ def read(self):
 
     def write(self, data):
         self.memory = data
+
+
+class EphemeralJSONStorage(Storage):
```
eugene-eeo: IMO you should subclass JSONStorage instead, that way you just need to override `write` and on a `read` you can do something like:

```
def read(self):
    if self.memory is not None:
        return self.memory
    return super().read(self)
```

msiemens:
Hey @imalento, thanks for the pull request :) What do you think about making this an extension, something like `tinydb-appengine`? If you make a small repository for this, I'll add it to the list of extensions :)

imalento:
Hi! Thank you for checking!
I also thought that this should be extension.
I'll make another small repository :smile:

msiemens:
Awesome, thanks 🙂

msiemens:
Just comment here when the repo is ready to be included in the extensions list 🙃

tinydb/storages.py
imalento: Thanks! I should have done so.

imalento:
Hi, I have made a repository. Thanks so much!
https://github.com/imalento/tinydb-appengine

msiemens:
Thanks, it's now in the list of extensions: https://tinydb.readthedocs.io/en/latest/extensions.html#tinydb-appengine 🙂

PR #270, opened 2019-06-27
Fix Query class to permit subclass creation

othalan:
Creating a subclass of the Query class in order to implement custom queries currently fails because the __getattr__ method creates a new instance of the Query class using a hard-coded name rather than the type of the current instance.  This can easily be fixed by using `type(self)()` to create the new class instance.

Creating a new subclass of Query is useful in order to simplify code scenarios where custom test functions are frequently used.

This pull request provides both a test and a fix for this problem.

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -366,17 +366,17 @@ def test_repr():
 
 
 def test_subclass():
+    # Test that a new query test method in a custom subclass is properly usable
     class MyQueryClass(Query):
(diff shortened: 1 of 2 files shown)


othalan:
The checks seem to be failing because of errors communicating with coveralls, not because of any issue with this change.

msiemens:
That's a good idea, thanks @othalan 🙂

PR #317, opened 2020-05-11
Replace setup.py calls with pip

### LINKED ISSUE #312: Install from source failed
I did:

```
git clone https://github.com/msiemens/tinydb/
cd tinydb
python setup.py install
```

I got 
"python: can't open file 'install': [Errno 2] No such file or directory"

And it's normal as I cannot find setup.py anywhere

hugovk:
Fixes #312.

It's generally discouraged to call `python setup.py`, even if it exists.

Running `pip install .` should install TinyDB from source. As TinyDB switched from `setup.py` to `pyproject.toml`, it doesn't include a `setup.py` file in the source tree any more.


msiemens:
Thanks ❤️

PR #352, opened 2020-11-05
feat: Make tinydb PEP 561 compatible

playpauseandstop:
[PEP 561](https://www.python.org/dev/peps/pep-0561/) defines a way to share type hints for Python packages.

As `tinydb` already provides type annotations for their modules, add last 2 pieces: `py.typed` file & PyPI classifier to allow `mypy` and other static analyzers use TinyDB type annotations.

---

First of all, thanks a lot for this beautiful piece of software! I'm enjoying use `tinydb` within my projects. 

However, as I recently restrict `mypy` config I start receiving errors such as,

```
src/app/contexts.py:5:1: error: Skipping analyzing 'tinydb': found module but no type hints or library stubs  [import]
src/app/contexts.py:5:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```

With that in mind I propose to make `tinydb` package PEP 561 compatible, so `mypy` and other static analyzer users may benefit from `tinydb` type hints.

Thanks

ps. BTW, to fix the `mypy` issue above I'm using next config,

```ini
[mypy-tinydb.*]
ignore_missing_imports = True
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
     "Operating System :: OS Independent",
+    "Typing :: Typed",
 ]
 
 packages = [


msiemens:
Thanks <3

msiemens:
This is now published in v4.3.0 🙂

playpauseandstop:
@msiemens 

Thanks a lot for merging this into `master` and making new 4.3.0 release, I appreciate it!

After upgrading to new release 95% of TinyDB code works smoothly with mypy and another 5% results in next issues:

- `error: Call to untyped function "TinyDB" in typed context  [no-untyped-call]`
- `error: Call to untyped function "Query" in typed context  [no-untyped-call]`
- `error: Argument 1 to "search" of "Table" has incompatible type "QueryInstance"; expected "Query"  [arg-type]`

Any of these issues are critical for me, so for now I just `type: ignore`'d them.

msiemens:
> After upgrading to new release 95% of TinyDB code works smoothly with mypy and another 5% results in next issues:

Good to hear that most of the code now passes. Do you have line numbers for the remaining errors?

playpauseandstop:
> Do you have line numbers for the remaining errors?

As the errors are in my code, I'm copy-pasting the code (with `# type: ignore` comments) below,

### Call to untyped function "TinyDB" in typed context

```python
    with TinyDB(settings.tinydb_path) as db:  # type: ignore[no-untyped-call]
        ...
```

### Call to untyped function "Query" in typed context

```python
Q = Query()  # type: ignore
```

### Argument 1 to "search" of "Table" has incompatible type "QueryInstance"; expected "Query"

```python
def search_stuff(
    db: TinyDB, *, external_uids: Tuple[uuid.UUID, ...]
) -> Tuple[StuffItem, ...]:
    return tuple(
        from_dict(StuffItem, item)
        for item in db.table(TABLE_STUFF).search(
            where(  # type: ignore[arg-type]
                "external_uid"
            ).one_of([str(item) for item in external_uids])
        )
    )
```

msiemens:
I looked into the type errors again and it turns out that not specifiying the return type on an `__init__` method makes it untyped:

> if `__init__` has no annotated arguments and no return type annotation, it is considered an untyped method:
>
> — https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods

I've pushed a fix which will be included in the next release.



> Argument 1 to "search" of "Table" has incompatible type "QueryInstance"; expected "Query"

Regarding this error it's more complicated as MyPy is technically correct: `Query` is a subclass of `QueryInstance`, not the other way around. Thus a function that expect a `Query` object won't accept `QueryInstance` objects. But in practice this works because `QueryInstance` has a `__call__` method to evaluate itself which is inherited to `Query`. I'll have to let this open for now

msiemens:
Just a quick note: TinyDB v4.5.0 now includes a fix for the Query error by providing a protocol for Query-like objects which both Query and QueryInstance implement 🙂

PR #377, opened 2021-03-01
Doc: Table missing in usage fixed

tozech:
The table of using document ids was missing in usage.rst due to a missing line break


msiemens:
Thanks ❤️

PR #389, opened 2021-04-07
Allow doc_id to be specified when using insert_multiple

waylonflinn:
Currently `doc_id` cannot be used with `insert_multiple` because no check is performed for the `Document` class. This PR adds the same code as used in `insert` to `insert_multiple`, thus allowing `doc_id` to specified for any item being added.

### DIFF
--- a/tinydb/table.py
+++ b/tinydb/table.py
@@ -187,9 +187,20 @@ class Table:
                 if not isinstance(document, Mapping):
                     raise ValueError('Document is not a Mapping')
 
-                # Get the document ID for this document and store it so we
-                # can return all document IDs later
-                doc_id = self._get_next_id()
+
+                # First, we get the document ID for the new document
+                if isinstance(document, Document):
+                    # For a `Document` object we use the specified ID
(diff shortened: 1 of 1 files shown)


msiemens:
Thanks for the pull request @waylonflinn! This looks good to me. Could you add a quick test case for this in `tests/test_tables.py`? Then I'd be ready to merge this

ianhinder:
It looks like this would allow you to insert documents with document IDs that already exist in the table.  `insert` currently checks that a supplied document ID does not already exist, as far as I can see.  Probably this PR should be changed to check that the provided document ID is not already in the table with a similar assertion to that used in `insert`.

benetherington:
@waylonflinn If you want to pass on these requested changes, I can work on this.

ianhinder:
This issue hasn't gone away, and the PR is still useful.  @waylonflinn - do you agree with my comment that an additional check should be added?

msiemens:
This has now been implemented by #455 which has been merged 🙂

PR #391, opened 2021-04-15
Switch to poetry-core

fabaff:
[`poetry-core`](https://github.com/python-poetry/poetry-core) is intended to be a light weight, fully compliant, self-contained package allowing PEP 517 compatible build frontends to build Poetry managed projects.

Using `poetry-core` allows distribution packages to depend only on the build backend.


msiemens:
Thanks for contributing! 🙂

Issue #394, opened 2021-05-05
Add upsert behavior for doc_ids (not just condition)


benetherington:
PR forthcoming, but I wanted to talk alternatives out first.

Upsert is one of my favorite tricks implemented by TinyDB. I'm currently working on an API where I'd really like to duplicate server db IDs in my local db. Fewer mistakes that way, I figure. Currently, Table.insert and Table.update handle doc_ids in very different ways. Insert doesn't take a doc_id argument, but instead requires a pre-built Document. Update, on the other hand, takes doc_ids as an argument. I suspect this has to do with the way Query() works, and that unifying this behavior would require a major version (maybe two! 😄)

The simplest solution would be to add a new method, maybe Table.upsert_by_id, but instead, I'd like to make upsert smarter to begin with. There are a few details to iron out, though.

**Upsert bridges an odd gap**, in that updating changes one or may records, but inserting only changes one. Upsert already wraps an insert-generated doc_id in a list to mimic a single updated doc_id's return value. This is good. An ID-aware upsert should accept doc_ids (plural). If it accepted doc_id (singular), it would be expected to return a single ID, not a list. That would be confusing.

**Simple is better.** When I started working on this, my first instinct was to break upsert into two branches of an if statement. One branch would handle cond but not doc_ids, the other would handle doc_ids but not cond. I added complexity by asserting only one argument was specified. I added further complexity when I had to loop over multiple inserts. I added EVEN FURTHER complexity when I tried to handle mixed existing and not-existing IDs. I stopped writing code when I was considering breaking off the doc_ids branch into a private method that upsert could call multiple times. It wasn't pretty, folks.

That's where I'm at right now, and I need to break off into a different task. When I come back, my plan is to keep the `if doc_ids is not None` branching structure, but A) group missing and existing IDs into two separate lists and handle them separately, B) change update_multiple to take a doc_ids argument and handle some of this logic, and maybe make it more useful to someone else down the line.

Does this all sound like a good plan of attack? Is the juice not worth the squeeze, and would Table.upsert_by_id be a better solution all round?

Thanks, folks, I really love this library, and I hope I can make it better. It fills a fantastic niche!

EDIT: AHHH should have checked open PRs first! There's already work to change update_multiple! Okay, I think that's really going to be the direction to take this.

benetherington:
Submitted PR #395.

Gardening and sleep lend clarity. Instead of adding a new argument to upsert, I followed the pattern established by insert and respected Document.doc_id.

[...]

Issue #371, opened 2021-02-05
Restructure and extend documentation


msiemens:
This is a note to myself to rework, restructure and extend the documentation based on the principles layed out in https://documentation.divio.com/introduction/ (and https://v3.vuejs.org/guide/contributing/writing-guide.html for that matter).

I considering adding a how-to section in the process. If anyone has ideas about what topics would be useful for short how-to guides/articles, feel free to comment on this issue 🙃

FallenSpaces:
Yes, constructing queries does not make much sense to me. 

Specifically, finding *all* docs with a given field that match or are greater than the given value.

From what I understand, you have to construct a query like this: ` Query().field.all(query | list)` ... but I don't understand how that works. How do I construct the `query` that goes in `.all()`? Reading the docs on this is confusing, no matter how much I go over it.

For example, I've also tried the following:

`docs = db.search(query.field.all(val))`

eugene-eeo:
probably writing extensions, with a little description of the internals (how to extend the code?).

vincentdavis:
It's been awhile since I worked with TinyDB and I was looking up how to use [db.upsert](https://tinydb.readthedocs.io/en/latest/usage.html#upserting-data). 

`>>> db.upsert({'name': 'John', 'logged-in': True}, User.name == 'John')`

This is confusing because `User.name` is defined near the top of the page, I was thinking to myself, is _User_ a table or another database? There are several places with this type of issue. It makes the otherwise awesome documentation a little difficult to use as a reference because the examples are not complete.

I would be happy to help with documentation or short how to guides or examples.
I am currently working to import data from Excel files doing a lot of db.upsert and field transformations. I could do some examples like this.

Point me in a direction and I can help some.
Thanks for TinyDB!

msiemens:
> This is confusing because User.name is defined near the top of the page, I was thinking to myself, is User a table or another database?

That's a good point. Basically the current documentation assumes that it is read from top to bottom. Instead we should split it into more subpages that have self-contained examples.

> I would be happy to help with documentation or short how to guides or examples.

That's great! 🙂 I think we should have an examples section that contains commonly used operations. If you like you could contribute your experience from working with TinyDB about what examples would be most helpful. I'm still not sure how to organize this best. Maybe I'll start a pull request where we could have discussions about these matters. Or maybe start collecting examples in the GitHub project wiki. What do you think?

msiemens:
Note to self: document using `upsert` with `Document`s

[...]

PR #395, opened 2021-05-06
Add doc_id handling to Table.Upsert() (#394)

benetherington:
Allows upsert to accept a Document with a doc_id attribute.

The cond argument is now optional if a Document (with doc_id) is passed.

### DIFF
--- a/tests/test_tinydb.py
+++ b/tests/test_tinydb.py
@@ -291,6 +291,28 @@ def test_upsert(db: TinyDB):
     assert db.upsert({'int': 9, 'char': 'x'}, where('char') == 'x') == [4]
     assert db.count(where('int') == 9) == 1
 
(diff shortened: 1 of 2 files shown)


msiemens:
Thanks for your work! I've thought about this for a bit and I think this is the best solution of adding `doc_id` support to `upsert` considering the differences of `doc_id` handling between `update` and `insert`.

The only thing that's left in my eyes is adding documentation about this in the usage documentation, but I think for now I'll just add a note about this in #371 so I'll remember this when reworking the documentation.

msiemens:
I'll try to get a release published in the next 1/2 weeks 🙂

benetherington:
I'd love to help with documentation, but it's so readable that I don't want to touch it. Thank you for your hard work!

Issue #401, opened 2021-06-09
default_table_name cannot be modified easily


A91y:
The Database always initiates with '_default' key value pair. 
So, A method should be added to modify dafault_table_name, so that a newbie can understand the code and update it accordingly.

A91y:
The Pull Request to do this modification is made [here](https://github.com/msiemens/tinydb/pull/402).

msiemens:
Hey @A91y, I think you also could use something like this:

```
TinyDB.default_table_name = 'something_else'
```

or alternatively use a subclass like this:

```
class MyDB(TinyDB):
    default_table_name = 'something_else'
```

Does one of these two approaches work for you?

msiemens:
Closing due to inactivity. Feel free to re-open it it's still unsolved for you.

PR #402, opened 2021-06-09
Added table_name method to TinyDB class

A91y:


msiemens:
As mentioned in #401 this already works by setting 

```python
TinyDB.default_table_name = 'something_else'
```

before creating the `TinyDB` instance 🙂

PR #410, opened 2021-06-28
Fix default_* cls variable usage in doc

laggardkernel:
`TinyDB.DEFAULT_*` class variables have been obsolete since 4.0. Drop and fix related usage example in the doc. Found this when reviewing the doc after reading the source code.


laggardkernel:
I forgot to do a search in the PRs and issues before opening this PR. Luckily, it seems this very PR solves the puzzle in #401 by accident.

msiemens:
Oh, that's a good catch! Thanks for helping out with this 🙂

PR #412, opened 2021-07-07
docs: add new upsert behavior information

MartinNiederl:
Added example of how to use the updated upsert behavior of pull request #395 to docs.


msiemens:
Thanks a lot 🙂

PR #424, opened 2021-08-06
Refactor Table.delete() to prefer argument doc_ids over cond

laggardkernel:
`Table.get(cond, doc_id)`, `.contains()`, `.update()`. These methods support both query condition and doc_id(s) as the filter. 
When query `cond` and `doc_id` are passed as params at the same, `doc_id` is preferred over `cond` and only `doc_id` is used.

Refactor `Table.delete()` to make it behave consistent with above methods.

### DIFF
--- a/tinydb/table.py
+++ b/tinydb/table.py
@@ -494,6 +494,25 @@ class Table:
         :param doc_ids: a list of document IDs
         :returns: a list containing the removed documents' ID
         """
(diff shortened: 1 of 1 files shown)


msiemens:
Good catch, thanks for contributing a fix! 🙂

msiemens:
This has now been released in TinyDB v4.5.2 🙃

PR #444, opened 2021-12-01
Update remaining reference to non-latest python version

rewritten:
In a couple of places the "main" (latest) version was still 3.9, this updates them to 3.10


msiemens:
Thanks! 🙂

PR #455, opened 2022-01-20
removed assert, added functionality of #389

MrPigss:
replaced assert statement with if and implemented a way to add documents with their existing doc_id using insert_multiple 
(see  #389 for original PR).
I added it since the original PR doesn't seem to active anymore.


msiemens:
Thanks @MrPigss and sorry for not responding earlier. I'm still relatively busy but hope to take a look at this next week or the week after that.

msiemens:
Thanks for your work @MrPigss! I'll try to get a release of TinyDB that contains this change published the next weekend or next week.

PR #460, opened 2022-01-30
Allow Table.get by doc_id to have O(1) behavior

richardsheridan:
`Table.get` is accidentally `O(len(Table))` because of the conversion of `doc_id` in `Table._read_table`. However, in the few places `_read_table` is used, there are opportunities to convert locally. Indeed, I discovered that `Table._get_next_id` was redundantly converting them already. Converting the user-input `doc_id` to string just before doing the `dict.get` call makes the overall lookup `O(1)`. ~~Also this change permits a bit of code sharing with `Table._update_table`.~~ (nope)

Of course, reading from the underlying storage is more like `O(nTables*avgLenTables)` in the standard case, so any speedup would only be realized when using `CachingMiddleware` or the like.

Looking forward to seeing if this passes the test suite! 😅

### DIFF
--- a/tinydb/table.py
+++ b/tinydb/table.py
@@ -289,7 +289,7 @@ class Table:
         if doc_id is not None:
             # Retrieve a document specified by its ID
             table = self._read_table()
-            raw_doc = table.get(doc_id, None)
+            raw_doc = table.get(str(doc_id), None)
 
             if raw_doc is None:
                 return None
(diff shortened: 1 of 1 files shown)


richardsheridan:
OK the problem seems to be inside actions/checkout. It is putting the head of master into the runner for testing instead of this PR's code, [see here](https://github.com/msiemens/tinydb/runs/5078346496?check_suite_focus=true#step:2:187).  And the resulting git log checkout is 967dbf9197b2943eb9bd91650f90bd9ab497bc12 which is an existing commit (currently master:HEAD).

This is super weird. The checkout command should be 

```bash
/usr/bin/git checkout --progress --force refs/remotes/pull/{PRnum}/merge
```

A little more research shows that this is an effect of triggering the CI runs from `pull_request_target`, which always runs against the PR merge base since it has elevated permissions to change the repo. Unfortunately I think it cannot be fixed until master reflects the change in trigger to `pull_request`. 

I've put this mess into another PR #462 and will reset this one to a simpler commit reflecting the original purpose, but hopefully not broken.

msiemens:
Thanks @richardsheridan, I expect to have a look at this soon 🙂

msiemens:
By the way, I'll try to get a release of TinyDB that contains this change published the next weekend or next week.

richardsheridan:
Awesome! I'm so happy to have contributed to the only database I may ever use :)

PR #462, opened 2022-02-05
Trigger CI on pull_request

richardsheridan:
pull_request_target always runs against the merge base commit because it provides elevated permissions. This means CI never actually runs on any pull request! See #460. I think the change will take effect after merging this commit to master, but I am not 100% sure.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -2,7 +2,7 @@ name: Python CI
 
 on:
   push: {}
-  pull_request_target:
+  pull_request:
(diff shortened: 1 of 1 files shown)


msiemens:
Thanks @richardsheridan. I'll take a look at this next week or the week after.

msiemens:
Thanks @richardsheridan! Seems like I got confused by how `pull_request_target` and `pull_request` work. I was under the impression that `pull_request_target` would still check out the Pull Request but run the workflow that is defined in the base repository. Seems like that was not correct.

Issue #467, opened 2022-02-19
Why not using os.rename to write file?


ostafen:
Hi, I noticed that the JsonStorage uses the write method to rewrite the entire file.
However if a crash occurs while the file is being written, the db could be damaged.
Why don't write to a temporary file and rename?
You can check what I mean in PR #468 :)

msiemens:
Hey! I've thought about this and I absolutely see the value of not damaging the database file when the process dies during a write operation. But I also see a cost in that creating a temporary file may not be something that application authors have considered when writing their applications. E.g. I can imagine this leading to some sort of weird permissions error where the running can't create new files in the directory of the database file.

As a compromise my thought was that this could be a new storage backend that is included with TinyDB, named something like `SafeJSONStorage` for example. This would also include documentation that describes the upsides and downsides of both this storage and the regular `JSONStorage` (similar to `yaml.safe_load` vs `yaml.load`).

What do you think?

ostafen:
Hi, Markus, thanks for your reply.
I think that crash recovery is something that each database, even the simplest, should deal with.
I can't imagine any application which persists data on disk where damaging the entire content of the database could be acceptable. If this is acceptable for your application, probably you application could store data in-memory.

Moreover, I think that if applications authors who are really concerned about performance, wouldn't probably choose TinyDB:
if we consider that the library doesn't focus on performance, the performance increase is tolerable.

However, a second solution is possible, if you want to optimize this aspect, which would be more performant, that is to say the double write technique. It avoids to create the temporary file at each write.
It consists in maintaining two files all the time, a main db file, and a temporary db file.
Each time you have to write the database on disk, you first truncate to zero length the temporary file and write the entire database to it.
At this point, you are sure that, if a crash happens, you can restore the database from the temporary file, so you can proceed to write data also to the main db file.
This means that, on application startup, you have to check that the main db file is not damaged. If this is the case, you can restore a valid version of the database from the temp file (which is for sure not damaged), otherwise you proceed normally.
I hope I have been able to explain myself.

Let me know what you think about this :-)

msiemens:
Thanks for your input, @ostafen! I haven't forgotten about this and will reply in the next few days! 🙂

msiemens:
Thanks again for your input, @ostafen!

I’ll admit that I’m really torn about this. Let me explain:

[...]

PR #468, opened 2022-02-21
Write to a temporary file and rename to avoid damage in case of crash

ostafen:
PR for #467

### DIFF
--- a/tinydb/storages.py
+++ b/tinydb/storages.py
@@ -6,6 +6,7 @@ implementations.
 import io
 import json
 import os
+import tempfile
 from abc import ABC, abstractmethod
 from typing import Dict, Any, Optional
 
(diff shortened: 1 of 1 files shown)


ostafen:
@msiemens Any update on this?

msiemens:
Hey @ostafen, I've been busy the last couple of weeks, I'll get back to this before the end of this week!

msiemens:
After merging this I've added Windows tests for TinyDB and it turns out that `os.rename` doesn't work on Windows when the destination file already exists (https://github.com/msiemens/tinydb/runs/6561237986?check_suite_focus=true) 🙈 I think I fixed it by adapting some code from https://github.com/untitaker/python-atomicwrites/ but the tests are still running.

It's not your fault that this went unnoticed, I think when you created your PR the CI configuration had an error so that it didn't run on Pull Requests. And even if it did, there were no Windows CI pipelines configured. So this one is on me 🙂 

I'll revert the changes for now and try to get this to work on a Windows machine so we can get this feature back in!

msiemens:
> think I fixed it by adapting some code from https://github.com/untitaker/python-atomicwrites/ but the tests are still running.

The tests have run now and they fail even with my workaround: https://github.com/msiemens/tinydb/runs/6561403728. I'll try to take a deeper look soon!

PR #477, opened 2022-06-28
Add return type for __hash__ function of QueryLike

masinag:
The implicit 'None' return type caused a warning from the linter


msiemens:
Thanks! 🙂

msiemens:
This has now finally been released 🎉  Sorry for the long delay!

PR #482, opened 2022-07-27
Atomic replace (second attempt)

### LINKED ISSUE #467: Why not using os.rename to write file?

richardsheridan:
Revamp of #468. Fixes #467.

I was in a position to fix this one because during the python-atomicwrites package fiasco I read somewhere that the author basically said people should be using `os.replace` anyway.  Other platform-dependent errors were fixed basically by closing file handles in the right order.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,7 +19,8 @@ def db(request):
         db_.drop_tables()
         db_.insert_multiple({'int': 1, 'char': c} for c in 'abc')
 
-        yield db_
+        with db_:
+            yield db_
 
 
 @pytest.fixture
--- a/tinydb/storages.py
+++ b/tinydb/storages.py
@@ -6,6 +6,7 @@ implementations.
 import io
 import json
 import os
+import tempfile
 from abc import ABC, abstractmethod
 from typing import Dict, Any, Optional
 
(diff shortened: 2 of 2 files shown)


richardsheridan:
CI problems seem unrelated, see https://github.com/msiemens/tinydb/runs/7544311658?check_suite_focus=true#step:5:231

richardsheridan:
A peculiar edge case of this implementation is that the storage will happily write to a "closed" handle. This is not triggered in the default case because a read is performed before writing, but it can be triggered by `CachingMiddleware`:

``` python
from tinydb import TinyDB, JSONStorage
db=TinyDB("test.json", storage=CachingMiddleware(JSONStorage))
db.insert({"test":1})
db.close()
db.insert({"test":2})
db.storage.flush()
db.storage.close()
```

I'll add a commit that checks `self._handle.closed` but finding this makes me wary of other edge cases. Maybe this feature should go in as a new "AtomicJSONStorage" and wait for some deprecation period and major version bump to be swapped in?

VermiIIi0n:
I don't think this is not triggered in the default case since that read operation is reading from the cache instead of the underlying storage.

How about making `CachingMiddleware` delete its cache after closing? (Which seems reasonable to me)

So that the next operation will read from the file and raise the exception.

Currently:
```Python
class CacheMiddelware(Middleware):
    def read(self):
        if self.cache is None:
            # Empty cache: read from the storage
            self.cache = self.storage.read()

        # Return the cached data
        return self.cache

def close(self):
        # Flush potentially unwritten data
        self.flush()

        # Let the storage clean up too
        self.storage.close()
```
->
```Python
def close(self):
        # Flush potentially unwritten data
        self.flush()
        self.cache = None  # Clear cache

        # Let the storage clean up too
        self.storage.close()
```

Issue #486, opened 2022-08-19
Query multiple docs by ID


pylipp:
As a follow-up to #329 , how can I query for multiple docs using their IDs? Like 

```
doc_ids = [1, 2, 5]
db.search(doc_ids=doc_ids)
```

`db.get(doc_id)` exists but only for a single document. I could iterate over the list of doc IDs yet this might not be as efficient.

msiemens:
That's interesting! Maybe we can make `db.get()` support a list of document IDs as `doc_id` (similar to `.remove()` does – altough it does have a `doc_ids` instead of `doc_id`). I'll have to think about that!

Atmden:
И как в итоге получить записи по списку doc_id?
Не получается использовать в search массив из doc_ids

patrick-nicodemus:
I would like to bump this feature request as I have a use-case for it.

keenborder786:
Hey , I will have a look into it and start working on it.

keenborder786:
@msiemens I have created a PR that inserts the given feature request. Please see #504

PR #504, opened 2023-01-23
Get Multiple Documents through get method of Table

keenborder786:
The feature augmentation in the table class is as per the requirement of #486. The user can now query multiple documents by specifying multiple doc_ids in form of a list in the get method of the Table class:

```python
db = TinyDB('/path/to/db.json')
db.insert({'int': 1, 'char': 'a'})
db.insert({'int': 1, 'char': 'b'})
db.get(doc_ids = [1,2]) # will return both the documents in form of list
```

### DIFF
--- a/tests/test_tinydb.py
+++ b/tests/test_tinydb.py
@@ -366,7 +366,10 @@ def test_get_ids(db: TinyDB):
     assert db.get(doc_id=el.doc_id) == el
     assert db.get(doc_id=float('NaN')) is None  # type: ignore
 
-
+def test_get_multiple_ids(db: TinyDB):
+    el = db.all()
+    assert db.get(doc_id=[x.doc_id for x in el]) == el
+    
 def test_get_invalid(db: TinyDB):
     with pytest.raises(RuntimeError):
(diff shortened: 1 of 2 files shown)


keenborder786:
Any updates?

msiemens:
Hey @keenborder786, sorry for not answering for so long! I was quite busy with work lately. All in all, this looks quite good, my only suggestion would be to not reuse the `doc_id` parameter but instead use another `doc_ids` parameter just like you mentioned in your PR description. This way using `doc_id` will not break when users use a document ID class other than `int` 🙂 Could you adapt your code in this manner?

keenborder786:
sure. I will have a look.

keenborder786:
@msiemens can you please see now?

msiemens:
Thanks for your contribution @keenborder786!

msiemens:
I'll try to prepare a release of TinyDB with this feature sometime in the next two weeks

msiemens:
Hey! I've now release TinyDB 4.8.0 that includes this feature 🙂

PR #505, opened 2023-01-23
Fix changelog link 404

okken:
Currently the changelog link on pypi results in 404.
PyPI page: https://pypi.org/project/tinydb/

Since people looking at the changelog from PyPI probably want released changes, the change is to point to the RTD changelog.
Another option is to point to the docs/changelog.rst, of course.


msiemens:
Thanks! 🙂

PR #508, opened 2023-02-05
feat: add query result limit

NafieAlhilaly:
#  🖐😃

#509 
### changes
Add `limit` option to `db.all()` and `db.search()` to limit the query result.
_this will also limit cached query result_

----
example
```python 
from tinydb import TinyDB, where


db = TinyDB("local.json")

# insert some data
for i in enumerate(range(0, 20)):
    db.insert({
        "id": i,
        "name": "awesome name"
    })

db.all()  # 20
db.all(limit=9)  # 9

ldb.search(where("id") > 1 and where("id") > 10)  # 9
db.search(where("id") > 1 and where("id") > 10, limit=3)  # 3

```

### DIFF
--- a/tinydb/table.py
+++ b/tinydb/table.py
@@ -2,7 +2,7 @@
 This module implements tables, the central place for accessing and manipulating
 data in TinyDB.
 """
-
+from itertools import islice
 from typing import (
     Callable,
     Dict,
@@ -218,10 +218,11 @@ class Table:
 
         return doc_ids
 
-    def all(self) -> List[Document]:
+    def all(self, limit: int = 0) -> List[Document]:
         """
         Get all documents stored in the table.
 
+        :param limit: limit the query result.
         :returns: a list with all documents.
         """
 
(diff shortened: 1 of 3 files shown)


tinydb/table.py
```diff
@@ -230,20 +231,27 @@ def all(self) -> List[Document]:
         # of all documents by using the ``list`` constructor to perform the
         # conversion.
 
+        if limit > 0:
+            return list(iter(self))[:limit]
```
ruancomelli, line 235: The `list(iter(self))` bit will store all elements of `self` in memory to throw possibly most of them away when slicing with `[:limit]`.
I suggest "slicing" the _iterator itself_ instead by making use of [`itertools.islice`](https://docs.python.org/3/library/itertools.html#itertools.islice):
```suggestion
            return list(itertools.islice(iter(self), limit))
```

NafieAlhilaly: Thank you.

tinydb/table.py
```diff
@@ -230,20 +231,27 @@ def all(self) -> List[Document]:
         # of all documents by using the ``list`` constructor to perform the
         # conversion.
 
+        if limit > 0:
+            return list(islice(iter(self), limit))
+
         return list(iter(self))
 
-    def search(self, cond: QueryLike) -> List[Document]:
+    def search(self, cond: QueryLike, limit: int = 0) -> List[Document]:
```
shpaker, line 239: Parameter "limit" without "offset" doesn't have any sense, in my view

PR #521, opened 2023-03-02
Use orjson instead native json to improve performance

gaetan1903:
orjson is a fast, correct JSON library for Python. 

it's written in rust with [maturin](https://github.com/PyO3/maturin), so it is fast 

A simple image to show difference of performance 

### WRITE DATA
save

### QUERY DATA
save2

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -415,6 +415,60 @@ files = [
 ]
 
 [[package]]
+name = "orjson"
+version = "3.8.7"
+description = "Fast, correct Python JSON library supporting dataclasses, datetimes, and numpy"
+category = "main"
+optional = false
+python-versions = ">=3.7"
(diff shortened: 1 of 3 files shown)


msiemens:
Thanks for your contribution, @gaetan1903! Unfortunately, I'll close this pull request. It's not that your suggestion or code is bad but rather that it's part of TinyDB's design (and its selling points) that it does not have any external dependencies. However, if you turn this into a new storage backend and release it on PyPI, I'll gladly link to it in the documentation and mention the performance benefits. That way users can decide if they prefer using Python's native `json` module or if they'd rather install (and potentially compile) a binary dependency in order to gain a performance increase. I hope you understand the reasoning behind this! If not, feel free to ask 🙂

PR #535, opened 2023-09-03
build(deps-dev): refactor unnecessary usage of `typing-extensions` and remove it from the dev-dependencies

ruancomelli:
Since this project requires Python>=3.8, we can always import `Protocol` directly from `typing`, making `typing-extensions` unnecessary.

This PR replaces `typing-extensions` with `typing` in the only place it was used and removes `typing-extensions` from the list of direct dev dependencies of the project.


msiemens:
Thank you 🙂

Issue #545, opened 2023-11-08
Document ID and existing unique value in the document


aptly-io:
When inserting documents and knowing a certain document property value (e.g. house-id) is unique, is there a way to tell tinydb to use that house-id as document ID?

Just found https://github.com/msiemens/tinydb/issues/351

The `document_id_class` is a class variable, it's for all tinydb's `Table` (looks not ideal at first sight).

Would this approach make sense:
When inserting a new Mapping, first make it a `Document(mapping, mapping.house-id)` (where the doc_id equal to the house-id).
However `Document` needs an `int` type `doc_it`. Should one also subclass `Document` to customize the `doc_id`'s type and change the `Table`'s `document_class`?
When creating a `Table` sub-class, one also has to create TinyDB sub-class to modify its class variable `table_class`?

Is there a more elegant approach that I'm overlooking?

aptly-io:
This might be a recipe for others when a str as doc_id:

```python

class _Document(dict):
    def __init__(self, value: Mapping, doc_id: str):
        super().__init__(value)
        self.doc_id = doc_id

class _Table(Table):
    document_class = _Document
    document_id_class = str
    def _get_next_id(self):
        raise NotImplementedError("doc_id should be self provided")

class _TinyDB(TinyDB):
    table_class = _Table


class Model:

    def __init__(self):
        self._db = _TinyDB("db.json")

```
There are some issue with table.py though when using _Document ...

msiemens:
I think your solution with a subclass is the best solution for your use-case (modulo the bug you found with the incorrect `isinstance` check).

PR #546, opened 2023-11-08
fix: isinstance with self.document_class iso Document

aptly-io:
I tried to sub-class `Document`, `Table` and `TinyDB` in order to redefine my own doc_id type
(see https://github.com/msiemens/tinydb/issues/545).
That failed because table.py checks isinstance against hard-coded `Document`.
By changing with `self.document_class` it worked.
Hence this PR

### DIFF
--- a/tinydb/table.py
+++ b/tinydb/table.py
@@ -147,7 +147,7 @@ class Table:
             raise ValueError('Document is not a Mapping')
 
         # First, we get the document ID for the new document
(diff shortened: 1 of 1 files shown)


msiemens:
Good point, thanks!

msiemens:
TinyDB v4.8.1 has been released which includes this bugfix! 🎉

PR #548, opened 2023-12-02
Test with Python 3.12 and add trove classifier

edgarrmondragon:


edgarrmondragon:
Tests are passing over on my fork, so let me know if there's interest in this.

ruancomelli (approved):
@msiemens these changes look correct. Can we get it merged? I'm also interested in having support for Python 3.12 :grinning:

edgarrmondragon:
> @msiemens these changes look correct. Can we get it merged? I'm also interested in having support for Python 3.12 😀

@ruancomelli FWIW Python 3.12 is already supported, this just _declares_ support via the PyPI trove classifier and adds 3.12 to the test matrix.

ruancomelli:
> @ruancomelli FWIW Python 3.12 is already supported, this just declares support via the PyPI trove classifier and adds 3.12 to the test matrix.

Hey @edgarrmondragon! Yes, I'm aware of that, but thanks for pointing it out!

I guess I meant "official support for Python 3.12", which is exactly what you did - without official support (including adding 3.12 to the test matrix), it would be unsafe to update `tinydb` in Python 3.12 projects since Python-3.12-incompatible changes could be introduced at any time... with official support and testing, this won't happen.

Additionally, some tools don't really like installing dependencies if official support for the project's Python version is not declared on PyPI, which makes this PR twice as valuable :)

edgarrmondragon:
Ping @msiemens :)

msiemens:
Thanks everyone!

PR #555, opened 2024-02-20
Fix broken example in "How to Extend TinyDB" docs

fvrlo:
Just a small misspelling fix for a page in the documentation.


msiemens:
Thanks!

PR #561, opened 2024-05-13
Add mention to no-op query in docs, tweak sentence

kinuax:
I wanted to build some dynamic queries and it took some time until I spotted the `noop` method. :sweat_smile: An explicit mention in the docs might help others in the same situation. Besides, a sentence seems to have two verbs.

Thanks and congrats for the project!


msiemens:
Thanks, @kinuax!

PR #571, opened 2024-10-06
Test with Python 3.13 and add trove classifier

edgarrmondragon:


edgarrmondragon:
`twine check` is failing due to this line

https://github.com/msiemens/tinydb/blob/bc93aad6cc21bd76e412ef166e97ecc902b4a166/README.rst#L2

https://github.com/pypa/readme_renderer/issues/304

Issue #581, opened 2024-11-18
[Improvement] Refactor too long function


higa-daniel:
After Sonar Cloud analysis, it has been detected a chance of code improvement. Basically, it simplifies the Conditional Logic of a function which is a little bit complex. 
For further information, please, check this out: https://sonarcloud.io/project/issues?cleanCodeAttributeCategories=ADAPTABLE&impactSeverities=HIGH&impactSoftwareQualities=MAINTAINABILITY&issueStatuses=OPEN%2CCONFIRMED&id=higa-daniel_tinydb&open=AZL5Xmrhs81F-WBnY1KF
I am opening a PR following the gitflow to be analyzed if it makes sense.
Thanks

PR #582, opened 2024-11-18
#581-refactor-too-long-function: improve function readbility

higa-daniel:
## Description
Improve function readability. Some nested conditional and repeated codes have been removed.

## Type of change
Improvement.

## How to test
Run unit tests without errors.

### DIFF
--- a/tinydb/table.py
+++ b/tinydb/table.py
@@ -383,131 +383,46 @@ class Table:
         :returns: a list containing the updated document's ID
         """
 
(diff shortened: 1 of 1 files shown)


msiemens:
Hey @higa-daniel, it seems like you completely removed the `update_multiple` method, is that right? Also, to me the complexity of the `update` method is not too much of a concern, as it has detailed line comments that explain the different parts of the function.

PR #583, opened 2024-11-19
Freeze Refactor - Diminish Cyclomatic Complexity

calebe-bertoluci:
Hello!
I would like to contribute to TinyDB by reducing the cyclomatic complexity of one of its methods.

I used the Codalyze Tool to statically analyze the code, and after extracting the result from all the software, it turns out that the freeze method has the highest cyclomatic complexity in TinyDB.
After the refactoring, it was possible to obtain a score of 6, for 3 points, on a scale of 0 to 10.

I hope you'll consider this change, and I'm excited to contribute to TinyDB.

Codalyze Tool: https://github.com/selcukusta/codalyze-rest-api

### DIFF
--- a/tinydb/utils.py
+++ b/tinydb/utils.py
@@ -145,15 +145,11 @@ def freeze(obj):
     """
     Freeze an object by making it immutable and thus hashable.
     """
-    if isinstance(obj, dict):
-        # Transform dicts into ``FrozenDict``s
-        return FrozenDict((k, freeze(v)) for k, v in obj.items())
-    elif isinstance(obj, list):
-        # Transform lists into tuples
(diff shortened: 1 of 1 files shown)


msiemens:
Hey @calebe-bertoluci, thanks for your PR! To be honest, to me the original looks cleaner even though it might have a higher cyclomatic complexity score. It is a simple if-else and not a Python dict that is used as a dispatch table (which is then immediately thrown away). Sure, the disptach table you're proposing has less conditionals, but I'd argue it's _harder_ to understand it compared to the original (it requires recognizing the code pattern in the first place, whereas the if-else is very easy to understand). What do you think?

msiemens:
Closing due to inactivity.

Issue #584, opened 2024-11-19
Add Edge Case and Special Scenario Tests


AlexanderPumaPucho:
Hi there 👋,  
I’m proposing to add a set of tests that cover edge cases and special scenarios to improve TinyDB’s robustness and reliability. These tests focus on situations like empty inputs, large data, special keys, nested structures, invalid configurations, and cache behavior, among others.

The goal is to ensure TinyDB handles uncommon scenarios gracefully and fails safely when necessary, without compromising data integrity. 🎯

### Test Summary

| **Category**                 | **What’s Being Tested**                                                                         | **Scenarios Covered**                                                                                                                                                          |
|------------------------------|------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| **Empty or Null Inputs**     | Validate how the database handles empty or `None` values.                                      | - Inserting empty dictionaries.<br>- Searching for nonexistent keys.<br>- Attempting to insert `None`.                                                                        |
| **Large Data Insertion**     | Test the database's ability to handle large documents.                                         | - Inserting very large strings and validating proper retrieval.                                                                                                               |
| **Special Keys and Values**  | Check support for unusual key-value combinations.                                              | - Keys with unusual characters.<br>- `None` values.<br>- Complex nested structures.                                                                                           |
| **Nonexistent Keys**         | Verify operations with keys that don’t exist.                                                 | - Searching, updating, or deleting nonexistent keys.                                                                                                                          |
| **Operations on Empty DB**   | Validate functionality when the database is empty.                                            | - Querying, counting, updating, or deleting when no data exists.                                                                                                              |
| **Invalid Configurations**   | Handle errors with incorrect configurations.                                                  | - Using an invalid storage class.                                                                                                                                             |
| **Transform Errors**         | Ensure failed transformations don’t corrupt data.                                             | - Simulating exceptions during data transformations.<br>- Ensuring partial transformations don’t affect database integrity.                                                  |
| **Cache Behavior**           | Validate cache consistency during reads and writes.                                           | - Testing read-after-write consistency.<br>- Multiple writes with a defined cache limit.                                                                                     |
| **JSON Storage and Persistence** | Validate JSON storage behavior and file size.                                             | - Ensuring data is saved and retrieved correctly.<br>- Verifying the file isn’t empty after saving data.<br>- Confirming file handles are properly closed.                    |

[...]

PR #586, opened 2024-11-19
Add test edge cases

AlexanderPumaPucho:
#584


msiemens:
Hey @AlexanderPumaPucho, thanks for your PR! Could you have a look at the test failures? It seems like they come from MyPy errors. If they are intentional (like using an incorrect document class definition), you can add line comments to turn off MyPy for a particular line: https://mypy.readthedocs.io/en/stable/error_codes.html

PR #592, opened 2025-05-14
docs(contributing): fix broken `GitHub Flow` url

ReenigneArcher:
`guides.github.com` doesn't exist anymore. I updated the URL for GitHub Flow to point to the reference in the docs instead.


msiemens:
Good catch, thanks!

msiemens:

PR #595, opened 2025-07-03
[PULL REQUEST] Getting faster performance

mejessie:
Switched from the built-in json library to #[orjson](https://github.com/ijl/orjson) to achieve faster serialization and JSON contents loading.


msiemens:
Thanks for the PR! I’m going to close this as not planned. Requiring users to install extra dependencies isn’t “tiny”, so this doesn’t fit the project goals.

Issue #544, opened 2023-10-18
is there any way to speed up deletion


fenchu:
My tindydb json file grows with 5GB per week if I do not delete.

Currently we just load the tinydb data.json and delete all internalids below a given threshold.

But the major problem is that we need to close tinydb handler to do this, this do not work well in a multiprocessing asyncio fastapi app.

I like to keep max 1000 entries in the list and delete everything below the 1000 highest.

Any guidelines on how to do this while keeping the app running would be great.

A suggestion I got was adding a timestamps (epoch) and delete any timestamps below the 1000 highest, but it bloats up the table.
and add extra logic.
Thanks

fenchu:
This can be obtained using db.max(), but it is slow
```
def keep_newest(key:str='jobid', maxlen:int=1000) -> Optional[List]:
    """ keep the newest maxlen entries in database """
    global db
    if not db:
        db = TinyDB(db_path)
    currlen = len(db.all())
    if currlen<=maxlen:
        #log.warning(f"database size is:{currlen} which is less than {maxlen} - no deletion")
        return False
    ids = []
    for d in db.all()[:currlen - maxlen + 1]:
        id = db.remove(where(key)==d[key])
        if id:
            ids.append(id)
        #log.info(f"removed {d} with index {id}")
    return ids
```

number of entries in database: 10000
number of entries in database: 999
deleting 9001 took 450.83sec

The json direct version is way faster:  1875 times faster?

```
def keep_newest_json(fname:str,  maxlen:int=1000, table:str='_default') -> Optional[List]:
    """ keep the newest maxlen entries in database """
    dat = None
    with open(fname, 'r', encoding='utf8') as FR:
        dat = json.load(FR)
    if table not in dat:
        log.fatal(f"table:{table} not found in dat:{list(dat.keys())}")
        return None
    currlen = len(dat[table].keys())
    if currlen<=maxlen:
        log.info(f"table:{table} has {currlen} entries, less than maxlen:{maxlen}")
        return None
    ids = []
    for id in list(dat[table].keys())[:currlen - maxlen + 1]:
        del dat[table][id]
        ids.append(id)
        #log.info(f"removed index {id} from {table}")
    with open(fname, 'w', encoding='utf8') as FW:
        FW.write(json.dumps(dat, indent=2, sort_keys=True))
    return ids
```
number of entries in database: 10000
number of entries in database: 999
deleting 9001 took 0.24sec

msiemens:
[...]

Issue #480, opened 2022-07-16
Suggesting Optimization for _update_table function


corradods:
I know TinyDB is not for high performance, but it seemed very strange to me that this code was taking forever, especially since there is no IO.
Example:
```
db = TinyDB(storage=tinydb.storages.MemoryStorage)
db.drop_tables()
for i in tqdm(range(1000000)):
    db.insert(tinydb.table.Document({'key': 'value'},doc_id=i))
```

The culprit seems to be these parts of the code in the function tinydb/table.py -> Table._update_table(...) :

```
        # Convert the document IDs to the document ID class.
        # This is required as the rest of TinyDB expects the document IDs
        # to be an instance of ``self.document_id_class`` but the storage
        # might convert dict keys to strings.
        table = {
            self.document_id_class(doc_id): doc
            for doc_id, doc in raw_table.items()
        }

#...#
        # Convert the document IDs back to strings.
        # This is required as some storages (most notably the JSON file format)
        # don't support IDs other than strings.
       tables[self.name] = {
            str(doc_id): doc
            for doc_id, doc in table.items()
        }
```

The comments are clear.  However, the overhead of this is huge, especially if i don't need it (e.g., i am not using a json file format, or i am already using strings as doc_ids type).
I extended Table for implementing a custom version of  _update_table function  that simply skip these conversion steps, and the example code can do the job in few seconds instead of being unfeasible.

another case of study:
```
CachingMiddleware.WRITE_CACHE_SIZE = 100000
Table.document_id_class = str
db = TinyDB('db.json', storage=CachingMiddleware(JSONStorage))

for i in tqdm(range(1000000)):
    db.insert(tinydb.table.Document({'key': 'value'},doc_id=str(i)))

```

So my question is, is it worth it?
And is it correct that this optimization is in the base table class?
Shouldn't it be borne by middlewares or storages which actually introduce the need for the optimization?

msiemens:
Hey @corradods,

I think this is a case where TinyDB tries to be flexible, but this flexibility comes at a cost. It is true that when you use the `MemoryStorage` or any other storage that supports arbitrary types for `dict` keys, this conversion is not needed. But, crucially, TinyDB does not know whether the storage will silently convert any `dict` key to string, as e.g. JSON does [1], or not.

[...]

PR #610, opened 2026-03-12
perf: add Bloom Filter for fast negative lookups

matheusvir:
## What was done

This PR introduces an in-memory Bloom Filter to optimize document lookup by `doc_id` in TinyDB.

By default, TinyDB performs `get(doc_id=...)` and `contains(doc_id=...)` by reading the full document table from storage, resulting in unnecessary I/O for nonexistent IDs. With this change, a Bloom Filter provides a fast probabilistic check that immediately discards IDs that are definitely not present, avoiding redundant storage access.

The Bloom Filter is enabled by default and requires no configuration or API changes.

## Implementation details

- Added a new module `bloom_filter.py` implementing a Bloom Filter with double hashing (MD5 + SHA-256).
- Modified `Table.__init__()` to accept optional parameters: `bloom_filter`, `bloom_expected_items`, and `bloom_fp_rate`.
- Uses lazy initialization: the filter is populated on the first `_read_table()` call, piggybacking on existing storage reads with zero extra I/O.
- Updated:
  - `get()` and `contains()` to skip storage access when the Bloom Filter confirms a `doc_id` is absent.
  - `insert()` and `insert_multiple()` to add new IDs to the filter.
  - `remove()` to rebuild the filter after deletions.
  - `truncate()` to clear the filter.
- No external dependencies were introduced.

## Complexity impact

| Operation | Without Bloom | With Bloom (miss) |
|---|---|---|
| `get(doc_id=N)` / `contains(doc_id=N)` | O(n) storage read | O(k) hash computations |

Where:
- *n* is the number of stored documents
- *k* is the number of hash functions (typically 6-10)

False positive rate is configurable (default: 1%), meaning at most 1% of lookups for absent IDs will still hit storage.

## Tests

Two new test files were added:

**`tests/test_bloom_filter.py`** — 7 classes, 27 tests covering the standalone `BloomFilter` in isolation: constructor validation, `add()`/`test()` correctness, zero-false-negative guarantee, hash determinism, `rebuild()`, serialization round-trips, and optimal parameter sizing.

**`tests/test_bloom_integration.py`** — 9 classes, 27 tests covering `BloomFilter` embedded inside `TinyDB.Table`, parametrized over both `MemoryStorage` and `JSONStorage`: lazy initialization, `insert`/`insert_multiple`, `get(doc_id=...)` and `contains()` short-circuit for absent IDs, `remove()` triggering a filter rebuild, `update()` consistency, full CRUD workflow, re-population on table reopen, and correct behavior when the filter is disabled.

All 260 tests pass with no regressions. Coverage: 95%.

---

## Performance

All benchmarks were executed inside Docker containers to isolate the runtime environment and eliminate host-specific variance from CPU scheduling, OS caching, and library versions.

### Methodology

- 50 total runs per dataset scale; first 10 (warmup) and last 10 (cooldown) discarded; 30 effective runs measured.
- Workload: 100 negative lookups per run — `get` and `contains` called with `doc_id` values guaranteed to be absent from the table.
- Storage: `JSONStorage`.
- Dataset scales: 1,000 / 10,000 / 50,000 documents.
- Timing: `time.perf_counter_ns()` with GC disabled during measurement.

### Results

| Scale | Baseline mean (ms) | Optimized mean (ms) | Improvement |
|---|---|---|---|
| 1k docs | 54.72 ± 1.87 | 1.81 ± 0.16 | **96.69%** |
| 10k docs | 787.03 ± 18.90 | 0.62 ± 0.08 | **99.92%** |
| 50k docs | 12,559.88 ± 1,349.95 | 1.29 ± 0.43 | **99.99%** |

Bloom Filter benchmark comparison

### Analysis

The gain scales dramatically with dataset size, which is the expected behavior for a Bloom Filter. Without the optimization, a negative lookup requires scanning the entire JSON file to confirm absence. With the filter, absence is determined in O(k) hash operations with no storage access.

At 50k documents, the baseline takes ~12.5 seconds per batch of 100 lookups; the optimized path takes ~1.3 ms — a reduction of roughly 9,700x. The standard deviation in optimized runs is also orders of magnitude smaller, indicating consistent and stable latency.

**Note:** as a probabilistic data structure, exact timings will vary across runs and machines. The no-false-negative guarantee is preserved: the filter never hides an existing document. False positives (at most 1% by default) simply fall back to the normal verification path.

### Reproducing the benchmark

The full benchmark infrastructure is available in the research repository at [matheusvir/eda-oss-performance](https://github.com/matheusvir/eda-oss-performance).

Relevant files:

- Dockerfile: [`setup/tinydb/Dockerfile`](https://github.com/matheusvir/eda-oss-performance/blob/main/setup/tinydb/Dockerfile)
- Benchmark script: [`experiments/tinydb/benchmark_bloom_filter.py`](https://github.com/matheusvir/eda-oss-performance/blob/main/experiments/tinydb/benchmark_bloom_filter.py)
- Runner script: [`experiments/tinydb/run_bloom_filter.sh`](https://github.com/matheusvir/eda-oss-performance/blob/main/experiments/tinydb/run_bloom_filter.sh)

To run:

```bash
# From the root of eda-oss-performance
bash experiments/tinydb/run_bloom_filter.sh
```

This builds the Docker image from `setup/tinydb/Dockerfile`, mounts the repository, and runs the benchmark script inside the container. Results are written to `results/tinydb/result_tinydb_bloom-filter.json`.

---

Feedback on filter sizing, the false positive rate default, and integration with other lookup paths is welcome.

---

Relates to #480 and #544.

### DIFF
--- a/tests/test_bloom_filter.py
+++ b/tests/test_bloom_filter.py
@@ -1,0 +1,262 @@
+"""
+Unit tests for the Bloom Filter implementation.
+
+This module tests the standalone BloomFilter class in isolation,
+covering initialization, add/test operations, false positive rates,
+hash determinism, rebuild, and serialization.
+"""
+
+import pytest
+
+from tinydb.bloom_filter import BloomFilter
+
+
+# ---------------------------------------------------------------------------
+# Initialization and parameter validation
+# ---------------------------------------------------------------------------
+
+class TestBloomFilterInit:
+    def test_default_parameters(self):
+        bf = BloomFilter()
+        assert bf.count == 0
+        assert bf.size > 0
+        assert bf.num_hashes > 0
+
+    def test_custom_parameters(self):
+        bf = BloomFilter(expected_items=500, false_positive_rate=0.05)
+        assert bf.count == 0
+        assert bf.size > 0
+        assert bf.num_hashes >= 1
+
+    def test_invalid_expected_items_zero(self):
+        with pytest.raises(ValueError, match='expected_items must be positive'):
+            BloomFilter(expected_items=0)
+
+    def test_invalid_expected_items_negative(self):
+        with pytest.raises(ValueError, match='expected_items must be positive'):
+            BloomFilter(expected_items=-10)
+
+    def test_invalid_fp_rate_zero(self):
+        with pytest.raises(ValueError, match='false_positive_rate must be between'):
+            BloomFilter(false_positive_rate=0.0)
+
+    def test_invalid_fp_rate_one(self):
+        with pytest.raises(ValueError, match='false_positive_rate must be between'):
+            BloomFilter(false_positive_rate=1.0)
+
+    def test_invalid_fp_rate_negative(self):
+        with pytest.raises(ValueError, match='false_positive_rate must be between'):
+            BloomFilter(false_positive_rate=-0.1)
+
+    def test_repr(self):
+        bf = BloomFilter(expected_items=100, false_positive_rate=0.01)
+        r = repr(bf)
+        assert 'BloomFilter' in r
+        assert 'size=' in r
+        assert 'hashes=' in r
+        assert 'count=' in r
+
+
+# ---------------------------------------------------------------------------
+# Add and test operations
+# ---------------------------------------------------------------------------
+
+class TestBloomFilterAddAndTest:
+    def test_add_and_test_single_item(self):
+        bf = BloomFilter(expected_items=100)
+        bf.add('hello')
+        assert bf.test('hello') is True
+
+    def test_add_and_test_multiple_items(self):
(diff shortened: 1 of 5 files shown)


msiemens:
Thanks for the thoughtful PR. The implementation, tests, and benchmarks all look solid.

I don’t think this is a fit for TinyDB itself, though. For me this falls a bit out of scope: it adds a fairly specialized optimization, plus extra internal complexity and configuration, to speed up one specific access pattern. Even if the numbers are good, that tradeoff doesn’t feel right for the core library.

I’d be very open to seeing this as an extension instead, where people who need this kind of optimization can opt into it explicitly without adding that complexity to TinyDB by default.

Issue #516, opened 2023-02-25
Modifying documents returned by `search` modifies the query cache


msiemens:
### Discussed in https://github.com/msiemens/tinydb/discussions/515

<div type='discussions-op-text'>

<sup>Originally posted by **sakam0to** February 23, 2023</sup>
A snippt codes
```
from tinydb import TinyDB, Query
db =  TinyDB("test.json")
db.insert({"id": 1})

q = Query()
records = db.search(q.id == 1)
records[0]["name"] = "N1"

db.search(q.id==1)
>>[{'id': 1, 'name': 'N1'}]
db.all()
>>[{'id': 1}]
```

You can see that modifications to the search function result affect the later search results.  But it doesn't change the records on the disk. Is it expected or not?</div>

ShubhamKaudewar:
We need to remove the cache read until we find a solution. Reading from db is ok than wrong data read from cache after modified.

msiemens:
Thinking about this more, I don't see any good solution. We can:

1. return deep copies of all results, but that degrades performance by >> 10x (see https://github.com/msiemens/tinydb/pull/617#issuecomment-5204091409),
2. make search results immutable, thereby breaking all current users who modify the search results, or
3. document this behavior and accept it as a quirk.

Given that option 1 is a _serious_ performance degradation and option 2 would require a v5 release just for this issue, I'm leaning toward optoin 3.

PR #617, opened 2026-06-24
docs: document that search() results share structure with the query cache

binggao1230:
## Update — converted to a docs-only PR

Per discussion below: a deep copy costs 150-4000x on the read hot path, and there's no fix that's both cheap and complete (a shallow top-level copy stops attribute reassignment but not nested mutation). Reverted the code change; the fix is now a docstring note on `Table.search()` plus a test that documents the actual sharing behavior (list is fresh, documents inside it are the cached objects).

## Original problem (#516)

`Table.search` stored and returned shallow copies of the document list, so a returned `Document` is the same object the cache holds. Mutating a key or a nested value in a returned document corrupts what a later identical search returns:

```python

db.insert({'name': 'Alice', 'tags': ['a', 'b']})

Q = Query()
results = db.search(Q.name == 'Alice')
results[0]['tags'].append('c')          # mutate the returned document

print(db.search(Q.name == 'Alice'))     # prints [{'name': 'Alice', 'tags': ['a', 'b', 'c']}]
```

## Tests

`test_query_cache_documents_are_shared_with_cache` documents the boundary: mutating the returned list is safe, mutating a document inside it is not. 219/219 pass.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -564,6 +564,12 @@ to the ``table(...)`` function:
 .. hint:: You can set ``cache_size`` to ``None`` to make the cache unlimited in
    size. Also, you can set ``cache_size`` to 0 to disable it.
 
+.. hint:: A repeated search with the same query is served from the cache.
+   The returned list is a fresh copy, but the documents inside it are the
+   same objects the cache holds — treat them as read-only. Mutating a
+   returned document (a key or a nested value) corrupts what a later
+   identical search returns.
+
 .. hint:: It's not possible to open the same table multiple times with different
    settings. After the first invocation, all the subsequent calls will return
    the same table with the same settings as the first one.
--- a/tests/test_tinydb.py
+++ b/tests/test_tinydb.py
@@ -697,6 +697,30 @@ def test_query_cache():
     assert db.search(query) == [{'name': 'foo', 'value': 42}]
 
 
+def test_query_cache_documents_are_shared_with_cache():
+    """search() returns a fresh list each call, but its documents are the
+    same objects the cache holds (see the docstring): the list itself can
+    be mutated freely, the documents inside it cannot."""
+    db = TinyDB(storage=MemoryStorage)
+    db.insert({'name': 'Alice', 'tags': ['a', 'b']})
+
+    query = where('name') == 'Alice'
+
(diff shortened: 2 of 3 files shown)


msiemens:
Thanks for looking into this! Unfortunately, this PR has a major flaw: it significantly degrades performance for *all* users of the cache:

| Scenario | master | PR #617 |
|---|---|---|
| cache hit, 10 docs matched | 0.4 µs | 62 µs (~150× slower) |
| cache hit, 1000 docs matched | 1.6 µs | 6,458 µs (~4,000× slower) |
| cache miss, 1000 docs matched | 502 µs | 6,994 µs (~14× slower) |

Ultimately, thinking about this again, I think there is no fix for #516 that does not introduce a performance penalty (deep copy) or break the API contract (make all results immutable). Given that my main issue was that the described behavior is *unexpected*, maybe it would be better to just properly document it. What do you think?

binggao1230:
Agreed, deep-copy isn't worth it here — the read path is the hot path and a 150-4000x hit for something meant to be a fast lookup is a bad trade for an edge case (mutating a returned result before the next identical search). I don't see a fix that's both cheap and complete: shallow-copying the top-level dict on read would stop attribute-reassignment corruption cheaply but not nested-value mutation, which is a half-fix that could still surprise someone. Happy to convert this into a docs-only PR noting that search() results share structure with the cache and must be treated as read-only — let me know if you'd rather write that note yourself.

msiemens:
A docs-only PR is fine with me

binggao1230:
Done — reverted the code change, kept a docstring note on `Table.search()` plus a test documenting the actual sharing boundary (fresh list, shared documents). 219/219 pass. Updated the PR title/body.

msiemens:
Thanks for reworking this! Could you also add a brief note to `docs/usage.rst` (section "Query Caching")? I assume most users will not regularly go into the API docs

binggao1230:
Done — added a note to the "Query Caching" section of `docs/usage.rst`, mirroring the `search()` docstring: cached searches return a fresh list, but the documents inside it are the cached objects, so treat them as read-only.

msiemens:
Thanks, @gaoflow!

Issue #517, opened 2023-02-25
Can't use lists/dicts as arguments to `Query.test`


msiemens:
### Discussed in https://github.com/msiemens/tinydb/discussions/514

<div type='discussions-op-text'>

<sup>Originally posted by **tgkuus** February 20, 2023</sup>
In the read the docs section Advanced queries>Custom test with parameters:
I noticed I can't have the arguments in the test function be lists or dicts. Is that by design or a bug?

Example of the issue. Note that the first error is at line 30. After the code is the error traceback I got.
```
from tinydb import TinyDB, Query

def test_function(val,arg1,arg2):
    if '0' == list(val.keys())[0]:
        return True
    else:
        return False
#make the db and add some data
x = TinyDB('questions.json')
x.truncate() #reset for testing
some_fake_data = [{'data':{xidx:xval}} for xidx,xval in enumerate(reversed(range(100)))]
print('1\n\t',some_fake_data)
x.insert_multiple(some_fake_data)
print('2\n\t',x.all())

#search normal - WORKS!
the_return = x.search(Query().data == {'0': 99})
print('3\n\t',the_return)

#search normal with test function args int and string - WORKS!
arg1=0
arg2='asd'
q=Query()
the_return = x.search(q.data.test(test_function,arg1,arg2))
print('4\n\t',the_return)

#search normal with test function args list and int - DOES NOT WORK
arg1=[0,1,2,3]
arg2=1
the_return = x.search(q.data.test(test_function,arg1,arg2))
print('5\n\t',the_return)

#search normal with test function args int and dict - DOES NOT WORK
arg1=1
arg2={'asd':1}
the_return = x.search(q.data.test(test_function,arg1,arg2))
print('6\n\t',the_return)

'''
Error Recieved
Traceback (most recent call last):
  File "...", line 31, in <module>
    the_return = x.search(q.data.test(test_function,arg1,arg2))
  File "...", line 245, in search
    cached_results = self._query_cache.get(cond)
  File "...", line 91, in get
    value = self.cache.get(key)
  File "...", line 98, in __hash__
    return hash(self._hash)
TypeError: unhashable type: 'dict'
'''

```</div>

PR #618, opened 2026-06-24
fix: freeze unhashable args in Query.test to prevent TypeError

binggao1230:
## Problem

Passing a `list` or `dict` as an extra argument to `Query().field.test()` raises
`TypeError: unhashable type: 'list'` (or `'dict'`) the first time TinyDB tries
to hash the resulting `QueryInstance` for the query cache.

```python
def in_set(val, allowed):
    return val in allowed

db.search(Query().x.test(in_set, [1, 2, 3]))  # TypeError: unhashable type: 'list'
```

Reported in https://github.com/msiemens/tinydb/discussions/514 (issue #517).

## Root cause

`Query.test` builds its cache key tuple as:

```python
('test', self._path, func, args)
```

`args` is a raw Python tuple that can contain unhashable objects (`list`,
`dict`).  Every other query method that accepts caller-supplied values
(`any`, `all`, `one_of`, `fragment`) already passes them through the
existing `freeze()` helper before building the hash tuple.  `test` was
the only method that did not.

## Fix

Apply `freeze()` to each element of `args` before composing the hash
tuple, matching the pattern used by the sibling methods:

```python
('test', self._path, func, tuple(freeze(arg) for arg in args))
```

`freeze()` recursively converts `list` → `tuple`, `dict` → `FrozenDict`,
and `set` → `frozenset`, so nested unhashable structures are handled too.
The test function still receives the original (unfrozen) `args` at
call-time; only the hash key is affected.

## Tests

Added `test_custom_with_unhashable_params` in `tests/test_queries.py`
covering list args, dict args, same-args hash stability, and different-args
hash distinctness.

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -222,6 +222,55 @@ def test_custom_with_params():
     assert hash(query)
 
 
+def test_custom_with_unhashable_params():
+    """Test that Query.test works when extra args contain lists or dicts."""
+
+    def in_allowed(value, allowed):
+        return value in allowed
+
+    # list argument must not raise TypeError when hashing the query
+    query = Query().val.test(in_allowed, [1, 2, 3])
+    assert query({'val': 1})
+    assert not query({'val': 4})
+    assert hash(query)
+
+    # dict argument must not raise TypeError when hashing the query
+    def has_key(value, mapping):
+        return value in mapping
+
+    query2 = Query().val.test(has_key, {'a': 1, 'b': 2})
+    assert query2({'val': 'a'})
+    assert not query2({'val': 'c'})
+    assert hash(query2)
+
+    # identical args produce the same hash (cache key correctness)
+    q_a = Query().val.test(in_allowed, [1, 2, 3])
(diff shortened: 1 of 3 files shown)


msiemens:
Thanks for the PR! Two quick thoughs:

1. Could we also add tuple handling into `freeze()` (`tinydb/utils.py:144-159`)? That way we don't get collisions when an arg is a `dict`
2. Could we also handle arbitrary unhashable args? Like numpy arrays etc. We could do `try: hash(hashval) except TypeError: hashval = None` to back out of the cache for cases like this

binggao1230:
Both done. `freeze()` now recurses into tuples too, so a nested dict/list inside an already-tuple arg is frozen instead of passed through raw. And `QueryInstance.__init__` now try/excepts `hash(hashval)`, falling back to `is_cacheable() == False` on any remaining `TypeError` (numpy arrays etc.) instead of crashing on first use. Added test cases for both; 216 pass.

msiemens:
Thanks!

msiemens:

Issue #575, opened 2024-11-11
Documentation Ideas - API Reference


mateustoin:
I was thinking of adding a Class Diagram and a simple explanation on how the classes correlate. Instead of having to read all of the API reference and the entire code to learn how the code works, people could start learning first on a high level explanation. This would be helpful for who wants to learn about the project and who wants to contribute later (that's my case).

The diagram could be something like this:
![image](https://github.com/user-attachments/assets/3c8953da-f745-461d-a013-e014d964b490)

Alongside with a high level explanation. What you think @msiemens? I hope to hear from you before I create a pull request about this.

Regards,
Mateus Antonio

msiemens:
Thanks for the ideas! I don’t really have capacity to take this on, but I'd be happy to review and merge a PR!

PR #620, opened 2026-07-01
docs: add architecture overview to API reference

uadhran:
Adds a high-level architecture section at the top of the API reference so readers can understand how TinyDB's core classes relate before diving into autodoc details.

## Changes

- New **Architecture Overview** section in `docs/api.rst` covering the three layers (core API, persistence, supporting components), typical data flow, and a link to `extend.rst` for contributors
- New class relationship diagram at `docs/_static/architecture.svg`

## Verification

- `uv run py.test -v --cov=tinydb` — not run (docs-only change)
- `sphinx-build -b html docs/ _build/html` — build succeeds; `architecture.svg` is copied into the output

Related: #575


msiemens:
Thanks @uadhran! I really like the idea but the SVG diagram seems to have some overlapping:

image

Could you fix that? I really like using Mermaid for diagrams like that, it takes care of the overlapping too. Ideally, you could commit the mermaid code along with the generated SVG

uadhran:
Fixed the overlapping diagram by regenerating it from Mermaid.

- Added `docs/_static/architecture.mmd` (source)
- Regenerated `docs/_static/architecture.svg` from that source
- Also rebased onto latest master

Regenerate with:
```bash
npx @mermaid-js/mermaid-cli -i docs/_static/architecture.mmd -o docs/_static/architecture.svg -b transparent
```

Issue #596, opened 2025-07-06
Bug in LRUCache implementation


epicblueprints:
``` python
class LRUCache(....):

def set(self, key: K, value: V):
  if self.cache.get(key):
    self.cache[key] = value
    self.cache.move_to_end(key, last=True)
  else:
    self.cache[key] = value

    # Check, if the cache is full and we have to remove old items
    # If the queue is of unlimited size, self.capacity is NaN and
    # x > NaN is always False in Python and the cache won't be cleared.
    if self.capacity is not None and self.length > self.capacity:
        self.cache.popitem(last=False)
```

The condition if self.cache.get(key): checks if the value is truthy, not if the key exists. This causes problems when Falsy values (0, False, "", [], None, etc.) are stored for a key.
When updating an existing key with a falsy value, it's treated as a new key (goes to else) and breaks LRU ordering.

A test case to demonstrate this bug. This test fails

``` python
def test_lru_cache_set_item():
    cache = LRUCache(capacity=3)
    cache["a"] = 0      # Falsy value!
    cache["b"] = 1
    cache["c"] = 2

    assert sorted(cache.lru) == ["a", "b", "c"]
    
    cache.set("a", 3)   # This should update "a" and move it to end
    cache.set("d", 4)   # This should evict "b"
    
    assert sorted(cache.lru) == ["a", "c", "d"]  # "b" should be gone but "a" is evicted in the current implementation
```

On the other end, the following test case passes - 
``` python

def test_lru_cache_set_item():
    cache = LRUCache(capacity=3)
    cache["a"] = 3
    cache["b"] = 1
    cache["c"] = 2

    assert sorted(cache.lru) == ["a", "b", "c"]

    cache.set("a", 0)
    cache.set("d", 4)

    assert sorted(cache.lru) == ["a", "c", "d"]
```

msiemens:
Closed in #597

PR #623, opened 2026-07-03
fix: LRUCache.get() and __getitem__() fail for cached None values

koteshyelamati:
## Problem

`LRUCache.get()` and `LRUCache.__getitem__()` both break when `None` is a legitimate cached value.

**`get()` bug:** uses `if value is not None:` to detect cache hits, so if a key maps to `None`, it skips the LRU promotion and returns the `default` instead of the cached `None`.

**`__getitem__()` bug:** calls `self.get(key)` and then checks `if value is None: raise KeyError`, so it raises `KeyError` even when the key exists with value `None`.

```python
cache = LRUCache(capacity=5)
cache['key'] = None

cache.get('key', 'missing')  # Returns 'missing' instead of None
cache['key']                  # Raises KeyError instead of returning None
```

## Root cause

Both methods use the cached value itself as a sentinel to detect cache misses. This is the same class of bug that was fixed in `set()` by PR #597.

## Fix

Replace the value-based checks with `key not in self.cache` membership tests, consistent with how `set()` was already fixed.

### DIFF
--- a/tinydb/utils.py
+++ b/tinydb/utils.py
@@ -78,9 +78,10 @@ class LRUCache(abc.MutableMapping, Generic[K, V]):
         del self.cache[key]
 
     def __getitem__(self, key) -> V:
-        value = self.get(key)
-        if value is None:
+        if key not in self.cache:
             raise KeyError(key)
+        value = self.cache[key]
+        self.cache.move_to_end(key, last=True)
 
         return value
 
@@ -88,14 +89,12 @@ class LRUCache(abc.MutableMapping, Generic[K, V]):
         return iter(self.cache)
 
     def get(self, key: K, default: Optional[D] = None) -> Optional[Union[V, D]]:
-        value = self.cache.get(key)
+        if key not in self.cache:
+            return default
+        value = self.cache[key]
+        self.cache.move_to_end(key, last=True)
 
-        if value is not None:
-            self.cache.move_to_end(key, last=True)
-
-            return value
-
-        return default
+        return value
 
     def set(self, key: K, value: V):
         if key in self.cache:


msiemens:
Thanks for the PR, @koteshyelamati! Could you add one or two tests for this?

uadhran:
Confirmed this still reproduces on current `master` (`4aa5311`, TinyDB 4.9.0) — not fixed by the recent docs/type PRs.

```python
from tinydb.utils import LRUCache

c = LRUCache(capacity=5)
c['k'] = None

assert 'k' in c.cache          # key is present
print(c.get('k', 'missing'))   # 'missing'  (expected None)
# c['k'] → KeyError            (expected None)
```

Root cause matches the PR: `get()` / `__getitem__()` still use `is not None` as a hit check (`tinydb/utils.py`), while `set()` already uses membership (`key in self.cache`) after #596/#597.

Also worth noting: `docs/changelog.rst` currently says falsy values like `0` or `None` are handled correctly in `LRUCache`, but `get`/`__getitem__` for stored `None` still fail — so this PR would also align the changelog claim with behavior.

### Suggested tests (if helpful)

Happy to help with tests if useful — something like:

```python
def test_lru_cache_none_value():
    cache = LRUCache(capacity=3)
    cache['key'] = None

    assert 'key' in cache.cache
    assert cache.get('key', 'missing') is None
    assert cache['key'] is None

    # LRU promotion still works for None hits
    cache['a'] = 1
    cache['b'] = 2
    cache.get('key')   # touch
    cache['c'] = 3     # would evict oldest if 'key' wasn't promoted
    assert cache.get('key') is None
```

(I don’t want to step on this PR — just adding verification + test ideas.)

R-zin (approved):

msiemens:
This has been handled by #635

PR #626, opened 2026-07-25
fix: adjust ID increment logic

jayanthmarupaka:
fixed the incremental logic by adding +2 instead of +1


msiemens:
With no explanation it is hard to know what you're trying to accomplish :) Closing this until further explanation has been provided

PR #621, opened 2026-07-01
fix: correct Table.update transform type hints

uadhran:
`Table.update` and `Table.update_multiple` annotated transform callables as
`Callable[[Mapping], None]`, but transforms mutate documents in place (e.g.
`tinydb.operations.delete` uses `del doc[field]`). `Mapping` has no
`__delitem__`, so the hints were inaccurate.

## Changes

- Change callable parameter to `Callable[[MutableMapping], None]` in
  `Table.update` and `Table.update_multiple`
- Add `test_update_custom_transform_callable` for custom transforms and
  `tinydb.operations` helpers

## Verification

- `uv run py.test -v --cov=tinydb` — 218 passed
- `uv run py.test -v --pycodestyle` — not run (pre-existing failures on master)

Related: #598

PR #627, opened 2026-07-27
Fix type hint for Table.update transform callback

jayasree-007:
## Summary

Update the type hint for the `Table.update` transform callback from
`Callable[[Mapping], None]` to `Callable[[MutableMapping], None]`.

The callback is expected to mutate the document in place. Existing usage,
including the `delete` transform, performs operations such as `del` on the
document, which requires a `MutableMapping` rather than a `Mapping`.

This change aligns the type annotation with the actual behavior.

## Testing

- Ran the full test suite.
- All tests passed.


msiemens:
Hey @jayasree-007, this has already been fixed in #621

PR #628, opened 2026-07-30
fix: use MutableMapping for Table.update callables

### LINKED ISSUE #598: Type hints for `Table.update` are inaccurate?
The type hints for `Table.update` say that if `fields` is `Callable`, then it must map `Mapping` to `None`. This probably either needs to be `Callable[[MutableMapping], None]` or `Callable[[Mapping], Mapping]`. 

For instance, the `delete` operation creates a "transform" function that deletes a key. But
```
def transform(doc: Mapping) -> None:
    del doc[field]
```
is invalid since `Mapping` doesn't have `__delitem__`: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes.

So either `doc` should be `MutableMapping`, or the operation should return the updated doc:
```
def transform(doc: Mapping) -> dict:
    doc = dict(doc)
    del doc[field]
    return doc
```

ltsyk:
## Summary
`Table.update` type for callable transforms used immutable `Mapping`, but transforms mutate the document.

## Change
Use `Callable[[MutableMapping], None]`.

Fixes #598


msiemens:
Hey @ltsyk, this has already been fixed in #621

PR #633, opened 2026-08-29
Don't reuse IDs in mixed insert_multiple batches

### LINKED ISSUE #629: insert_multiple can silently overwrite documents when mixing Document and dict
### Description

`Table.insert()` resets `_next_id` when inserting a `Document` with an explicit ID, so later auto-assigned IDs do not collide. `Table.insert_multiple()` does **not** reset `_next_id` on the `Document` path, and auto-ID allocation inside the same batch still calls `_get_next_id()`, which reads **storage** (not the in-progress updater table).

As a result, a mixed `insert_multiple([Document(...), plain_dict, ...])` can:

1. Return **duplicate IDs** in the result list
2. **Silently overwrite** a document that was just inserted earlier in the same batch

### Minimal reproduction

```python
from tinydb import TinyDB
from tinydb.table import Document

ids = db.insert_multiple([
    Document({'a': 1}, doc_id=1),
    {'b': 2},
])

print(ids)      # [1, 1]  — duplicate IDs
print(db.all()) # [{'b': 2}]  — first document lost
```

Second path (stale `_next_id` after a normal `insert`):

```python
db.insert({'x': 1})  # id=1; _next_id becomes 2
ids = db.insert_multiple([
    Document({'a': 1}, doc_id=2),
    {'b': 2},
])
print(ids)      # [2, 2]
print(db.all()) # [{'x': 1}, {'b': 2}]  — Document({'a': 1}, 2) overwritten
```

Contrast with `insert()`, which already resets `_next_id`:

```python
db.insert(Document({'a': 1}, doc_id=5))
db.insert({'b': 2})
# documents have ids 5 and 6 — no collision
```

### Code pointers

- `insert()` resets `_next_id` for `Document`: `tinydb/table.py` (around the `isinstance(document, self.document_class)` branch)
- `insert_multiple()` Document branch stores the ID but does **not** reset `_next_id`, then the dict branch calls `_get_next_id()`
- `_get_next_id()` uses `self._next_id` or `_read_table()` from storage — not the updater’s in-memory `table`

I’d like maintainer input on the intended contract. Reasonable options:

1. Make `insert_multiple` match `insert`: reset `_next_id = None` when a `Document` is inserted, and/or allocate auto-IDs from the in-progress `table` (max existing + 1) so IDs cannot collide within the batch
2. Raise `ValueError` on collision within the batch (same as duplicate explicit `Document` IDs already do)
3. Document that mixed `Document` + auto-ID dicts in one `insert_multiple` call is unsupported

Silent overwrite feels like the least acceptable outcome.

### Suggested tests

- `insert_multiple([Document({...}, 1), {...}])` on empty table must not return `[1, 1]` / must not drop the first document
- After `insert(...)`, `insert_multiple([Document({...}, next_id), {...}])` must not overwrite the explicit Document
- Keep existing `test_insert_multiple_with_doc_ids` behavior for all-Document batches

### Environment

- Python 3.11

gyanu2507:
`insert()` resets `_next_id` after a Document with an explicit ID. `insert_multiple()` didn't, and auto-ids in the same batch were still allocated from storage, so `[Document(..., 1), {...}]` overwrote id 1.

Fixes #629


msiemens:
Thanks for the PR! Could you have a look at the failing tests?

```
FAILED tinydb/table.py::mypy - 217: error: Incompatible types in assignment (expression has type "int", variable has type "None")  [assignment]
221: error: Incompatible types in assignment (expression has type "int", variable has type "None")  [assignment]
```

gyanu2507:
The mypy failures were `self._next_id = <int>` after `if self._next_id is None`, which had narrowed the attribute to None.

`_next_id` is typed as `Optional[int]` now. The next id is computed on a local, then written back, then `_get_next_id()` runs as before.

msiemens:
Thanks!

PR #634, opened 2026-08-29
Reject CachingMiddleware reads and writes after close

### LINKED ISSUE #631: CachingMiddleware: operations after close() still succeed in memory
### Description

After `TinyDB.close()` (or leaving a context manager), using `CachingMiddleware(JSONStorage)` still allows `insert` / `all` etc. to succeed against the **in-memory cache**. The underlying storage handle is already closed, so those writes are **not durable**.

Plain `JSONStorage` fails on post-close I/O (closed file). The middleware path is inconsistent and easy to mistake for a successful write.

### Minimal reproduction

```python
import os, tempfile
from tinydb import TinyDB
from tinydb.storages import JSONStorage

fd, path = tempfile.mkstemp(suffix='.json')
os.close(fd)

db = TinyDB(path, storage=CachingMiddleware(JSONStorage))
db.insert({'a': 1})
db.close()

# Appears to work:
new_id = db.insert({'b': 2})
print(new_id)    # 2
print(db.all())  # [{'a': 1}, {'b': 2}]

# But on-disk file still only has the pre-close data:
print(open(path).read())
# {"_default": {"1": {"a": 1}}}

os.unlink(path)
```

Note: `CachingMiddleware.close()` correctly `flush()`es before closing storage, so the first insert is on disk. The **post-close** insert only mutates `self.cache` and is lost when the process exits (unless something flushes again — which would hit a closed handle).

### Code pointers

- `CachingMiddleware.close()` flushes then closes storage, but does **not** clear `self.cache` or mark the middleware closed (`tinydb/middlewares.py`)
- `TinyDB.close()` closes storage and clears table cache, but does not prevent further table operations (`tinydb/database.py`)
- Post-close `CachingMiddleware.write()` only updates `self.cache` until a flush threshold

Possible directions (no preference assumed):

1. Raise a clear error on read/write after close (fail fast)
2. Clear `CachingMiddleware.cache` on close and reject further ops
3. Document that the DB must not be used after `close()`, and that middleware may still appear to accept writes

### Suggested tests

- After `close()` with `CachingMiddleware(JSONStorage)`, `insert` / `all` should either raise **or** (if documented) not claim durability — ideally assert the on-disk file is unchanged by post-close inserts
- Compare behavior with plain `JSONStorage` for consistency

### Environment

- Python 3.11

gyanu2507:
After `close()`, CachingMiddleware kept serving the in-memory cache, so inserts looked fine but never hit the closed file. Reads/writes now fail once the middleware is closed.

Fixes #631


msiemens:
Thanks for the PR!

One thing about the test: `test_caching_rejects_use_after_close` was added in the middle of `test_caching_json_write`. The end of that test (`del db` and the part that reopens the database and checks `db.all()`) now belongs to the new test, so `test_caching_json_write` no longer checks that the data was written when the context manager exits. Could you move that part back to the end of `test_caching_json_write` and give the new test its own ending?

gyanu2507:
`test_caching_json_write` again ends with `del db` and the reopen/`db.all()` check.

`test_caching_rejects_use_after_close` keeps its own reopen so a flush-on-close still lands on disk.

msiemens:
Thanks!

PR #635, opened 2026-09-05
Fix LRUCache treating a stored None as a missing key

### LINKED ISSUE #632: LRUCache treats stored None values as absent (KeyError from __getitem__, wrong get() result)
## Summary

`LRUCache` (exported via `tinydb.utils.__all__` and documented as mimicking a dictionary) breaks `MutableMapping` semantics for entries whose value is `None`: `cache[key] = None` succeeds and `key in cache` is `True`, but `cache[key]` raises `KeyError`, and `cache.get(key, default)` returns `default` even though the key exists. Additionally, reading an existing `None` entry never refreshes its LRU position.

*This report is based on static analysis of master; I have not executed the code.*

## Location

- File: `tinydb/utils.py`
- Class: `LRUCache`
- Methods: `__getitem__`, `get`

```python
def __getitem__(self, key) -> V:
    value = self.get(key)
    if value is None:
        raise KeyError(key)          # treats a *stored* None as absent

    return value

def get(self, key: K, default: Optional[D] = None) -> Optional[Union[V, D]]:
    value = self.cache.get(key)

    if value is not None:            # falsy/None values skip move_to_end
        self.cache.move_to_end(key, last=True)
        return value

    return default                   # returns default even if key exists
```

## Problem

The implementation conflates "value is `None`" with "key absent". `OrderedDict.get(key)` correctly distinguishes them (returning the sentinel only when missing), but this wrapper discards that distinction. Consequences:

1. `cache[key] = None; cache[key]` → `KeyError`, violating the `Mapping` contract (`__getitem__` must succeed for contained keys; note `__contains__` reports `True`).
2. `cache.get(key, default)` returns `default` for an existing `None` entry.
3. An existing `None` entry is never moved to the MRU end, so its LRU bookkeeping is stale relative to accesses.

For contrast, a sentinel-based check (`value = self.cache.get(key, _MISSING); if value is _MISSING: ...`) would preserve the contract.

## Trigger / Reproduction

```python
from tinydb.utils import LRUCache

c = LRUCache(capacity=3)
c['a'] = None
assert 'a' in c     # True
c['a']              # raises KeyError — should return None
c.get('a', 'dflt')  # returns 'dflt' — should return None
```

(Static-analysis finding — derived from the code above.)

## Expected Behavior

`LRUCache` honors `collections.abc.MutableMapping` semantics: contained keys are retrievable regardless of value, and only genuinely absent keys produce `KeyError`/`default`.

## Actual Behavior

`None`-valued entries behave as if absent for retrieval, while `__contains__`, `__len__`, and eviction accounting still count them — an internally inconsistent view of the cache.

## Impact

`LRUCache` is part of TinyDB's public API surface (`__all__`) and is documented for general use ("This class acts as a dictionary"). Any consumer caching `None`-valued results hits silent misbehavior. Internally TinyDB itself only stores lists (never `None`), which is presumably why this went unnoticed — but the class is reusable utility code.

## Suggested Direction

Use a private sentinel for misses in `get()`, and drive `__getitem__`/move-to-end off the sentinel rather than `value is None`.

afonsojanu:
Closes #632.

`LRUCache.__getitem__` and `get()` both use `value is not None` as their presence check, which conflates "the key holds None" with "the key isn't here at all". So caching a None result gives you a KeyError on read, `get()` silently hands back the default instead of None, and the entry never gets bumped to the MRU end on access even though it's sitting right there in the OrderedDict.

Fixed by checking against a module-level sentinel instead of None, both in `__getitem__` (now looks the key up directly against `self.cache` rather than going through `get()`) and in `get()` itself. Added `test_lru_cache_none_value` covering the KeyError case, the `get()` default case, and that a None entry still moves to the front of the LRU order on access.

Ran the full suite plus a fresh mypy pass over `tinydb/` afterward, both clean.


msiemens:
Thanks for the PR!

PR #643, opened 2026-09-19
fix: coerce doc_ids in update/remove like get

### LINKED ISSUE #639: get(doc_id='1') works but update/remove with doc_ids=['1'] silently no-op
### Description

`Table.get` coerces IDs with `str(...)` when looking up storage keys:

```python
raw_doc = table.get(str(doc_id), None)
# and for doc_ids:
doc_ids_set = set(str(doc_id) for doc_id in doc_ids)
```

(`tinydb/table.py`)

`Table.update` / `Table.remove` with `doc_ids=` test membership against the in-memory table dict, whose keys are `document_id_class` values (ints by default) — **without** the same coercion:

```python
updated_ids.extend(
    doc_id for doc_id in requested_ids if doc_id in table
)
```

So `get(doc_id='1')` finds a document, but `update(..., doc_ids=['1'])` / `remove(doc_ids=['1'])` return `[]` and change nothing.

### Minimal reproduction

```python
from tinydb import TinyDB

db.insert({'x': 1})

print(db.get(doc_id='1'))
# {'x': 1}

print(db.update({'x': 2}, doc_ids=['1']))
# []
print(db.all())
# [{'x': 1}]  — unchanged

print(db.remove(doc_ids=['1']))
# []
print(len(db))
# 1  — still there

# int IDs work as expected:
print(db.update({'x': 2}, doc_ids=[1]))
# [1]
```

Possible directions:

1. Coerce `doc_ids` in `update`/`remove` the same way `get` does (via `document_id_class` / consistent key form)
2. Reject non-`document_id_class` values with a clear error
3. Document that `get` accepts stringified IDs but `update`/`remove` require int IDs

Silent no-op on writes while reads succeed is the surprising part.

### Suggested tests

- After insert, `get(doc_id='1')` vs `update(..., doc_ids=['1'])` / `remove(doc_ids=['1'])` — assert consistent success or consistent rejection per chosen option
- Keep existing int-ID and missing-ID (#591) behavior

### Environment

- Related consistency work: #591 (missing int IDs)

uadhran:
## Summary

`get(doc_id='1')` finds a document, but `update` / `remove` with `doc_ids=['1']` returned `[]` and changed nothing because table keys are ints.

Per #639 option (1): coerce IDs with `document_id_class` in `update` / `remove`, matching `get`’s ability to accept stringified IDs.

Fixes #639

## Test plan

- [x] New `test_doc_ids_string_coercion`
- [x] Existing update/remove ID tests still pass
- [x] Manual issue repro now updates and removes with `doc_ids=['1']`


uadhran:
Fixed the Python 3.14 mypy failures: annotations for `get`/`update`/`remove` now allow `int | str` IDs (matching the coercion behavior), `get(doc_id=...)` goes through `document_id_class`, and the regression test narrows `Document | None` before indexing.

